PR #18, opened 2018-10-04
Removed value_field from config.yaml

epeterson12:
Removed `value_field `from config.yaml. Now, the `value_field ` in images_to_samples.py is taken from the prep csv file


mpelchat04:
Tests in CI didn't pass. Here's the error:
image
Related to this [line in the code](https://github.com/NRCan/geo-deep-learning/blob/6276842438e6d54694d2991fd906e39609074758/utils.py#L171) where 'info' is not defined in the function.

PR #63, opened 2019-02-25
Added validation in training for the number of classes in classification tasks

### LINKED ISSUE #57: Add a Validation for the Number of Classes in Classification Tasks
In train_model.py, there isn't any verification that the number of classes used for training in classification tasks is the same as the parameter given in the config file. The model is built based on the the config file's value for num_classes

If the wrong number of classes is used for training, it results in meaningless outputs from the model and ```IndexError: list index out of range ``` during classification.

We need to add a check to avoid running meaningless trainings.

epeterson12:
Closes #57


epeterson12:
Wrong version of code commited

Issue #84, opened 2019-06-12
Receptive field calculated from max pooling layers


mpelchat04:
The receptive field of a model is currently calculated from the number of max pooling layers in the model's architecture. There are some limitations to this approach:  
- Max pooling layers can be replaced with strides in conv layers, therefore augmenting the model's receptive field.  
- Max pooling of size > 2x2 are not currently handled.  

Depending of the architecture, calculating the receptive field of a model can be tricky/impossible. Therefore, it would be best to remove the function to calculate the [receptive field](https://github.com/NRCan/geo-deep-learning/blob/master/models/model_choice.py#L10) and simply replace it with:
a) a constant value, parametrizable from the config file; OR
b) a constant value deduced from the size of the input patch (e.g. 10% in height/width)

mpelchat04:
Closed in #89

Issue #11, opened 2018-09-27
Manage discontinuities when classifying


ymoisan:
When classifying images, we can see linear artifacts in the final output due to the moving window used to assign classes being smaller than the image to be classified. An example: 

![image](https://user-images.githubusercontent.com/7468595/46153678-9ef02080-c241-11e8-96a2-c68da52b618f.png)
 
The solution would be to manage the moving window overlap and get the mean value of the prediction for overlapping regions. See [Audebert et al., (2016)](https://arxiv.org/abs/1609.06846)

mpelchat04:
# Problem
The reason for these artifacts is that we need to tile the image, in order to fit it into memory and classify it. For a good pixel classification, the model rely on the context of the pixel. At the border of the tile, the model doesn't have any context to classify the pixel, giving results as shown above. 
# Solution
The idea is to find the number of pixel needed for context and write only the pixels with enough context, in the tile. To ensure that each pixel i correctly classified, each tile will overlap it's neighbors by the number of pixel required for context.  
The number of pixel required for context is depending on the depth of the model. Every time there is a resampling in the model, we add more context to the classification. Given that resampling are kernel of size 2x2, the number of pixels used for context can be obtained by 2**n where n is the depth of the model (number of times resample is done).

PR #89, opened 2019-07-09
Fixup attempt for overlap artefacts in segm inference

plstcharles:
Untested, merged from old code! (I don't have the necessary data on my side to run it anymore)


mpelchat04:
Great job, fixes the artefacts.  
Better fixup for #11   
Thanks for your contribution!

mpelchat04:
Closes #84 - No need for receptive field calculation anymore. One only has to specify an overlap % and all predictions are averaged, during inference.

PR #107, opened 2019-10-28
Delete models parameter section, automatically manage output dir, etc.

remtav:
- conf/config.yaml: remove models parameter section
- inference.py: optionally input a directory with .tif files instead of a .csv
- inference.py: add debug mode.
- utils/model_choice.py: returns state_dict instead of path to state_dict
- utils/utils.py: add tolerance in assigning GPU to task, if some memory is used (ex.: 10%) and some usage existing (ex.: 10%)
- images_to_samples.py, inference.py and train_model.py: create output directory automatically with pathlib

### DIFF
--- a/train_model.py
+++ b/train_model.py
@@ -237,9 +237,9 @@ def set_hyperparameters(params, model, checkpoint):
 
     # optional hyperparameters. Set to None if not in config file
     class_weights = torch.tensor(params['training']['class_weights']) if params['training']['class_weights'] else None
-    if class_weights:
+    if params['training']['class_weights']:
         verify_weights(params['global']['num_classes'], class_weights)
-    ignore_index = params['training']['ignore_index'] if params['training']['ignore_index'] else None
+    ignore_index = params['training']['ignore_index'] if params['training']['ignore_index'] else -100
 
     # Loss function
     criterion = MultiClassCriterion(loss_type=params['training']['loss_fn'], ignore_index=ignore_index, weight=class_weights)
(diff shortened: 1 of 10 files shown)


train_model.py
```diff
@@ -1,9 +1,8 @@
-from pathlib import Path
-
 import torch
 # import torch should be first. Unclear issue, mentioned here: https://github.com/pytorch/pytorch/issues/2083
 import argparse
 import os
+from pathlib import Path # TODO Use Path instead of os where possible. Better cross-platform compatibility
```
mpelchat04: Good thinking!

utils/utils.py
```diff
@@ -219,7 +202,7 @@ def read_csv(csv_file_name, inference=False):
         return sorted(list_values, key=lambda k: k['dataset'])
 
 
-def get_device_ids(number_requested):
+def get_device_ids(number_requested): #FIXME if some memory is used on a GPU before call to this function, the GPU will be excluded.
```
mpelchat04: `if some memory is used on a GPU before call to this function, the GPU will be excluded`... That's exactly what this function does. Since we don't know in advance how much memory is needed for the process we want to push to the GPU, we exclude those with data on it. There might be a better way to solve this, tough. Any suggestions?

remtav: Actually, my note is unclear. What I meant is:
Sometimes, a bit of GPU memory is used by a python script (for what reason, I don't know...) but you still want to use that GPU. Therefore I modified code to include GPU even if, say, 10% of it's memory is used. I also added a threshold for the %usage for the same reason. 

The function was fine to begin with, I just raised the threshold (harcoded) to include partly used GPUs.

PR #123, opened 2020-04-02
NIR integration 

victorlazio109:
- Modified colormap.csv to re-order classes and assign appropriate colours
- Added code to modify deeplabv3 weights to support 4 bands
- Added a simple script to crawl directory for images and corresponding gpkg files, writes to csv file
- Modified padding strategy while dynamically allocating images to val set.


images_to_samples.py
```diff
@@ -366,21 +417,28 @@ def main(params):
...
                 if params['sample']['mask_reference']:
                     np_label_raster = mask_image(np_input_image, np_label_raster)
 
                 if info['dataset'] == 'trn':
                     out_file = trn_hdf5
-                elif info['dataset'] == 'val':
-                    out_file = val_hdf5
+                    val_file = val_hdf5
+                # elif info['dataset'] == 'val':
+                #     out_file = val_hdf5
```
CharlesAuthier, line 441: Delete the comment.

models/model_choice.py
```diff
@@ -57,10 +57,18 @@ def net(net_params, num_channels, inference=False):
...
-            model = models.segmentation.deeplabv3_resnet101(pretrained=False, progress=True,
-                                                            num_classes=num_channels, aux_loss=None)
+            # assert num_bands==3, 'Edit torchvision scripts segmentation.py and resnet.py to build deeplabv3_resnet ' \
+            #                      'with more or less than 3 bands'
+            print('Testing with 4 bands')
+            import numpy as np
+            import torch.nn as nn
+            model = models.segmentation.deeplabv3_resnet101(pretrained=True, progress=True, aux_loss=None)
+            conv1 = model.backbone._modules['conv1'].weight.detach().numpy()
+            depth = np.random.uniform(low=-1, high=1, size=(64, 1, 7, 7))
```
CharlesAuthier, line 67: Are you introducing noise on a fifth band here?

models/common.py
```diff
@@ -52,4 +54,72 @@ def forward(self, x):
...
 
+class DeepLabHead(nn.Sequential):
+    def __init__(self, in_channels, num_classes):
+        super(DeepLabHead, self).__init__(
+            ASPP(in_channels, [12, 24, 36]),
+            nn.Conv2d(256, 256, 3, padding=1, bias=False),
+            nn.BatchNorm2d(256),
+            nn.ReLU(),
+            nn.Conv2d(256, num_classes, 1),
+            nn.Tanh()
```
mpelchat04, line 65: I'm curious as to why exactly is this new code is exactly the same as in the original [DeepLabV3 code](https://github.com/pytorch/vision/blob/master/torchvision/models/segmentation/deeplabv3.py), except for the `nn.Tanh()` line?

models/model_choice.py
```diff
@@ -57,10 +57,18 @@ def net(net_params, num_channels, inference=False):
...
         except:
-            assert num_bands==3, 'Edit torchvision scripts segmentation.py and resnet.py to build deeplabv3_resnet ' \
-                                 'with more or less than 3 bands'
-            model = models.segmentation.deeplabv3_resnet101(pretrained=False, progress=True,
-                                                            num_classes=num_channels, aux_loss=None)
+            # assert num_bands==3, 'Edit torchvision scripts segmentation.py and resnet.py to build deeplabv3_resnet ' \
+            #                      'with more or less than 3 bands'
+            print('Testing with 4 bands')
+            import numpy as np
+            import torch.nn as nn
```
mpelchat04, line 64: Add those imports at the beginning of the script.

train_segmentation.py
```diff
@@ -306,15 +306,15 @@ def main(params, config_path):
...
-    try: # For HPC when device 0 not available. Error: Cuda invalid device ordinal.
+    try:  # For HPC when device 0 not available. Error: Cuda invalid device ordinal.
         model.to(device)
     except RuntimeError:
         warnings.warn(f"Unable to use device. Trying device 0...\n")
         device = torch.device(f'cuda:0' if torch.cuda.is_available() and lst_device_ids else 'cpu')
         model.to(device)
+    model, criterion, optimizer, lr_scheduler = set_hyperparameters(params, num_classes_corrected, model, checkpoint)
+
+    criterion = criterion.to(device)
```
mpelchat04, line 317: I don't know if it has changed or not in pytorch but as I recall, we had to initialize the optimizer before pushing the model to cuda device. @remtav do you recall this as well?  
Here's an explanation: https://discuss.pytorch.org/t/effect-of-calling-model-cuda-after-constructing-an-optimizer/15165

images_to_samples.py
```diff
@@ -86,11 +98,11 @@ def check_sampling_dict():
             if type(int(key)) == int:
                 pass
                 # if type(value) == int:
-                    # pass
+                # pass
```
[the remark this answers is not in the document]
victorlazio109: Yes, would resolve

images_to_samples.py
```diff
@@ -119,18 +131,43 @@ def class_proportion(target):
...
+            val = True
+            _samples_file = val_sample_file
+        append_to_dataset(_samples_file["sat_img"], data)
+        append_to_dataset(_samples_file["map_img"], target)
+        append_to_dataset(_samples_file["meta_idx"], metadata_idx)
+
+    else:
+        append_to_dataset(samples_file["sat_img"], data)
+        append_to_dataset(samples_file["map_img"], target)
+        append_to_dataset(samples_file["meta_idx"], metadata_idx)
```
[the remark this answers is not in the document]
victorlazio109: That certainly looks better, I would test the code out

models/common.py
victorlazio109: Yes you are right very similar code. The reasoning behind it is FineTuning. Since the actual pretrained weights is being called. Some of the modules and layers of DeepLabV3 has to be modified for our use case.
1.) First modification is on the first layer to take an additional channel. An additional weight dimesion was initialized and appended to the exisiting conv filters effectively changing the first conv size from 64, 3, 7, 7 --> 64, 4, 7, 7
2.) Second modification is on the classifier layer which is why that piece of code exists. For the pretrained weights the classifier layer explicitily supports 21 classes (trained on Coco Set) whereby we are predictiing on 4 classes. So based on this piece of code we are overwriting the classifier layer to support any number of classes (4 classes in our own case). The idea is too retain the pretrained weights for all the other layers and modify first layer and last layers for our use case. Hence Finetuning for our use case.

models/model_choice.py
victorlazio109: No, I am introducing a fourth band with its weights initialzed randomly. Can be interpreted as noise that is learnable to support an additional fourth band.

models/model_choice.py
victorlazio109: Would resolve

train_segmentation.py
victorlazio109: I encountered an error when initalizing the optimizer before pushing the model to cuda. Construting the initializer uses the model parameters as seen from the code. However when it is constructed before the model is pushed for instance to a cuda device(GPU) a mismatch occurs between the parameters ( model initialized on CPU vs Model pushed to Cuda device).  Apparently, this problem does not occur for the SGD optimizer, but for ADAM and AdaGrad the problem is encountered. Here are some official docs and discussion on the issue:
https://pytorch.org/docs/master/optim.html
https://discuss.pytorch.org/t/effect-of-calling-model-cuda-after-constructing-an-optimizer/15165

images_to_samples.py
victorlazio109: Would resolve

images_to_samples.py
```diff
@@ -7,6 +7,7 @@
 import numpy as np
 import warnings
 import rasterio
+import random
```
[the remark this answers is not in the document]
victorlazio109: Would test np.random out. Certainly would want to reduce imports

models/common.py
victorlazio109: P.S. I would change the activation layer to Relu. Relu is much better for multi class problems

images_to_samples.py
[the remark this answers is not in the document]
victorlazio109: Reviewing it now Charles, like you said it commented on this branch and master. Is everyone okay for me to remove it?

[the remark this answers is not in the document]
CharlesAuthier: i am :p

images_to_samples.py
```diff
@@ -149,58 +174,66 @@ def samples_preparation(in_img_array, label_array, sample_size, overlap, samples
...
             for column in range(0, w, dist_samples):
-                data = (pad_in_img_array[row:row + sample_size, column:column + sample_size, :])
-                target = np.squeeze(pad_label_array[row:row + sample_size, column:column + sample_size, :], axis=2)
-
+                data = (in_img_array[row:row + sample_size, column:column + sample_size, :])
+                target = np.squeeze(label_array[row:row + sample_size, column:column + sample_size, :], axis=2)
+                data_row = data.shape[0]
+                data_col = data.shape[1]
+                if data_row < sample_size or data_col < sample_size:
+                    data = pad_diff(data, data_row, data_col, sample_size)
```
remtav, line 206: Cool!

train_segmentation.py
remtav: Well the doc seems clear: better to construct the optimizer after pushing model to GPU. I seem to recall this part of the code was a little tricky to move and we had to do everything in a particular order. But anyways, I does seem more logical with the provided doc to push model before constructing optimizer...

images_to_samples.py
```diff
@@ -54,6 +53,18 @@ def mask_image(arrayA, arrayB):
     return ma_array
 
 
+def pad_diff(arr, w, h, arr_shape):
```
remtav, line 56: Would be a good habit (for all of us!) to add docstrings to our functions.

images_to_samples.py
```diff
@@ -119,7 +124,15 @@ def class_proportion(target):
     return False
 
 
-def compute_classes(samples_file, data, target, metadata_idx, dict_classes):
+def compute_classes(dataset, samples_file, val_sample_file, data, target, metadata_idx, dict_classes):
```
remtav, line 127: Would also add docstrings here.

images_to_samples.py
```diff
@@ -119,7 +124,15 @@ def class_proportion(target):
...
+    val = False
+    if dataset == 'trn':
+        random_val = np.random.randint(1, 100)
+        if random_val > 5:
```
remtav, line 131: So 5% is going to validation? Wouldn't it be best not to hardcode this? At least, I would document it in the README.

images_to_samples.py
victorlazio109: Good Idea!

images_to_samples.py
victorlazio109: You are right, best not to hardcode. Would add a val_percent param to the config file to be set by user. thanks

models/model_choice.py
```diff
@@ -82,21 +89,6 @@ def net(net_params, num_channels, inference=False):
...
-            coco_model = models.segmentation.fcn_resnet101(pretrained=True, progress=True, num_classes=21, aux_loss=None)
-        checkpoint = coco_model.state_dict()
-        # Place entire state_dict inside 'model' key for compatibility with the rest of GDL workflow
-        temp_checkpoint = {}
-        temp_checkpoint['model'] = {k: v for k, v in checkpoint.items()}
-        del coco_model, checkpoint
-        checkpoint = temp_checkpoint
-    elif pretrained:
-        warnings.warn(f'No pretrained checkpoint found for {model_name}.')
-        checkpoint = None
```
remtav, line 99: Much cleaner without this :)

train_segmentation.py
```diff
@@ -146,9 +147,12 @@ def create_dataloader(data_path, batch_size, task, num_devices, params, samples_
     num_workers = num_devices * 4 if num_devices > 1 else 4
 
     # Shuffle must be set to True.
-    trn_dataloader = DataLoader(trn_dataset, batch_size=batch_size, num_workers=num_workers, shuffle=True, drop_last=True)
-    val_dataloader = DataLoader(val_dataset, batch_size=batch_size, num_workers=num_workers, shuffle=False)
-    tst_dataloader = DataLoader(tst_dataset, batch_size=batch_size, num_workers=num_workers, shuffle=False) if num_samples['tst'] > 0 else None
+    trn_dataloader = DataLoader(trn_dataset, batch_size=batch_size, num_workers=num_workers, shuffle=True,
+                                drop_last=True)
+    val_dataloader = DataLoader(val_dataset, batch_size=batch_size, num_workers=num_workers, shuffle=False,
+                                drop_last=True)
```
remtav, line 153: We preferred not dropping last items (drop_last=True) for val and test sets since shuffle is False and batch sizes are usually 1 (right?). Therefore, it is not necessary to drop last batch if smaller than batch size. Moreover, it causes for those last samples to be consistently ignored (always the same ones, since shuffle is False). @mpelchat04 That was the reason, right? I think you were the one to bring it up.

I would suggest keeping drop_last = False for val and test dataloaders.

utils/augmentation.py
```diff
@@ -60,10 +64,28 @@ def __call__(self, sample):
...
+
 class ToTensorTarget(object):
     """Convert ndarrays in sample to Tensors."""
+    def __init__(self, num_classes):
+        self.num_classes = num_classes
 
     def __call__(self, sample):
         sat_img = np.float32(np.transpose(sample['sat_img'], (2, 0, 1)))
         map_img = np.int64(sample['map_img'])
+        map_img[map_img > self.num_classes] = 0
```
remtav, line 90: Not sure I understand the purpose of this line. What if map_img carries ignore_index values that are bigger than num_classes (e.g. 255, which is the case in our current dataset)

utils/visualization.py
```diff
@@ -112,6 +114,8 @@ def vis(params, input, output, vis_path, sample_num=0, label=None, dataset='', e
                 new_ignore_index = 255
                 label[label == ignore_index] = new_ignore_index  # Convert all pixels with ignore_index values to 255 to make sure it is last in order of values.
 
+    if params['training']['normalization']['mean'] and params['training']['normalization']['std']:
+        input = unnormalize(input_img=input, mean=mean, std=std)
```
remtav, line 118: Good thinking

utils/visualization.py
```diff
@@ -97,6 +97,8 @@ def vis(params, input, output, vis_path, sample_num=0, label=None, dataset='', e
     heatmaps = get_key_def('heatmaps', params['visualization'], False)
     grid = get_key_def('grid', params['visualization'], False)
     ignore_index = get_key_def('ignore_index', params['training'], -1)
+    mean = get_key_def('mean', params['training']['normalization'])
+    std = get_key_def('std', params['training']['normalization'])
```
remtav, line 101: Will it cause a bug if these parameters are missing from the yaml config?

utils/utils.py
```diff
@@ -62,36 +62,7 @@ def load_from_checkpoint(checkpoint, model, optimizer=None, inference=False):
...
     """
-    # Corrects exception with test loop. Problem with loading generic checkpoint into DataParallel model
-    # https://github.com/bearpaw/pytorch-classification/issues/27
-    # https://discuss.pytorch.org/t/solved-keyerror-unexpected-key-module-encoder-embedding-weight-in-state-dict/1686/3
-    if isinstance(model, nn.DataParallel) and not list(checkpoint['model'].keys())[0].startswith('module'):
-        new_state_dict = model.state_dict().copy()
-        new_state_dict['model'] = {'module.'+k: v for k, v in checkpoint['model'].items()}    # Very flimsy
-        del checkpoint
-        checkpoint = {}
-        checkpoint['model'] = new_state_dict['model']
```
remtav, line 73: These lines were useful for a specific use case: training saves a generic non-DataParallel model. Then we load that model during test time into an instanciated DataParallel model. Throws exception if key in pretrained non-DataParallel dict don't start with "module.". I have a feeling these lines should stay, but I'm willing to try without, then revert if necessary.

utils/utils.py
```diff
@@ -62,36 +62,7 @@ def load_from_checkpoint(checkpoint, model, optimizer=None, inference=False):
...
-                    mismatched_layers.append(mismatch_layer)
-            if inference:
-                assert len(mismatched_layers) == 0, f"Layers {mismatched_layers} mismatch with current model. During " \
-                                                    f"inference, layers shouldn't be chopped."
-            chopped_checkpt = chop_layer(checkpoint['model'], layer_names=mismatched_layers)
-            # overwrite entries in the existing state dict
-            model.load_state_dict(chopped_checkpt, strict=False)
-        except RuntimeError as error:
-            raise RuntimeError(error)
-
```
remtav, line 94: Bon débarras as we'd say in spanish!

utils/augmentation.py
```diff
@@ -60,10 +64,28 @@ def __call__(self, sample):
...
+        self.std = std
+
+    def __call__(self, sample):
+        if self.mean or self.std != []:
+            sat_img = (sample['sat_img'] - self.mean) / self.std
+            map_img = sample['map_img']
+            return {'sat_img': sat_img, 'map_img': map_img}
+        else:
+            return sample
+
```
remtav, line 80: Wouhou! :)

utils/augmentation.py
```diff
@@ -26,7 +26,11 @@ def compose_transforms(params, dataset):
             lst_trans.append(RandomRotationTarget(limit=params['training']['augmentation']['rotate_limit'],
                                                   prob=params['training']['augmentation']['rotate_prob']))
 
-    lst_trans.append(ToTensorTarget())
+    if params['training']['normalization']['mean'] and params['training']['normalization']['std']:
```
remtav, line 29: Would recommend the systematic use of get_key_def as a way to add robustness to our scripts. May not be necessary here, but just making sure it's a conscious decision. 
For example, get_key_def can save us bugs if value for specific parameter or the parameter itself is omitted in yaml. in that case, it assigns a default value (if added as parameter)

utils/augmentation.py
```diff
@@ -26,7 +26,11 @@ def compose_transforms(params, dataset):
...
+        lst_trans.append(Normalize(mean=params['training']['normalization']['mean'],
+                                   std=params['training']['normalization']['std']))
```
remtav, line 31: This will change slightly in light of my upcoming PR, since it will include Scaling. Should we force scaling to 0-1 range if we're normalizing?

train_segmentation.py
victorlazio109: The defined batch_size in the config yaml is retained for trn,val and test dataloaders especially when training on GPU. I came across issues when the batch_size divides the number of samples and there is a remainder. This issue does not arise when the batch size occurs in even chunks of the total number of samples. It has nothing to do with the shuffle param.

utils/augmentation.py
victorlazio109: Am still up for setting the scale dynamically. Sometimes scalling from [-1-1] may be beneficial.

utils/augmentation.py
victorlazio109: it was hacky fix, that solved the problem with using other loss functions such as cross_entropy loss_fn

utils/visualization.py
victorlazio109: No, although it is not tested extensively. I may be wrong. If you come across any bugs please report!

PR #141, opened 2020-06-25
Bugfixes and minor refactoring

remtav:


images_to_samples.py
```diff
@@ -450,22 +452,14 @@ def main(params):
                     raise ValueError(f"Dataset value must be trn or tst. Provided value is {info['dataset']}")
                 val_file = val_hdf5
 
-                # TODO: save histogram to metadata, then use it at radiometric trimming, if chosen.
-                metadata = raster.meta
-                metadata['name'] = raster.name
-                metadata['csv_info'] = info
+                metadata = add_metadata_from_raster_to_sample(sat_img_arr=np_input_image,
```
victorlazio109, line 455: Can we wrap metadata functionality in a condition? This would resolve some errors that may arise when metadata for a raster does not exist or if it is incorrect.

inference.py
```diff
@@ -240,7 +238,14 @@ def main(params):
     img_dir_or_csv = params['inference']['img_dir_or_csv_file']
 
     default_working_folder = Path(params['inference']['state_dict_path']).parent.joinpath(f'inference_{num_bands}bands')
-    working_folder = Path(get_key_def('working_folder', params['inference'], default_working_folder)) # TODO: remove working_folder parameter in all templates
+    working_folder = get_key_def('working_folder', params['inference'], None)
```
victorlazio109, line 241: I think we may need to dicuss custom working folders vs automatic creation of folders. For instance, it currently unintuitive  to share our train, val, tst Hdf5s since it exists in a folder that was automatically created. When devloper A runs training the logs and output are saved in the same directory as the datasets similarly for developer B. At this point it  becomes difficult to separate the training runs for each developer. I believe outputs and logs should be saved in a user defined path (custom working folder).

utils/augmentation.py
```diff
@@ -60,6 +60,13 @@ def compose_transforms(params, dataset, type='', ignore_index=None):
...
     if type == 'totensor':
+        if not dataset == 'trn' and random_radiom_trim_range:  # Contrast stretching at eval. Use mean of provided range
+            RadiometricTrim.input_checker(random_radiom_trim_range)  # Assert range is number or 2 element sequence
+            if isinstance(random_radiom_trim_range, numbers.Number):
+                trim_at_eval = random_radiom_trim_range
+            else:
+                trim_at_eval = round((random_radiom_trim_range[-1] - random_radiom_trim_range[0]) / 2, 1)
+            lst_trans.append(RadiometricTrim(random_range=[trim_at_eval, trim_at_eval]))
         if scale:
             lst_trans.append(Scale(scale))  # TODO: assert coherence with below normalization
```
victorlazio109, line 71: We need to sort a uniform scaling function, so as not to introduce bugs while using radiometric trim.

utils/verifications.py
```diff
@@ -7,99 +7,7 @@
 from tqdm import tqdm
 
 from utils.CreateDataset import MetaSegmentationDataset
-from utils.geoutils import lst_ids
-from utils.utils import get_key_recursive
-
-
-def is_valid_geom(geom):
```
victorlazio109, line 14: was this funtion moved somewhere else?

utils/utils.py
```diff
@@ -331,4 +319,31 @@ def read_csv(csv_file_name):
         list_values = sorted(list_values, key=lambda k: k['dataset'])
     except TypeError:
         list_values
-    return list_values
\ No newline at end of file
+    return list_values
+
+
+def add_metadata_from_raster_to_sample(sat_img_arr: np.ndarray,
```
victorlazio109, line 325: worried about the effect of processing raster with no metadata. This function has be used in a condition statement to account for raster without metadata or incorrect meta.

models/model_choice.py
```diff
@@ -57,17 +57,21 @@ def net(net_params, num_channels, inference=False):
         assert (num_bands == 3 or num_bands == 4), msg
         if num_bands == 3:
             print('Finetuning pretrained deeplabv3 with 3 bands')
-            model = models.segmentation.deeplabv3_resnet101(pretrained=pretrained, progress=True, aux_loss=None)
-            model.classifier = common.DeepLabHead(2048, num_channels)
+            model = models.segmentation.deeplabv3_resnet101(pretrained=True, progress=True)
```
CharlesAuthier, line 60: We force people to use the pretrained weight?

models/model_choice.py
```diff
@@ -57,17 +57,21 @@ def net(net_params, num_channels, inference=False):
...
+            classifier = list(model.classifier.children())
+            model.classifier = nn.Sequential(*classifier[:-1])
+            model.classifier.add_module('4', nn.Conv2d(classifier[-1].in_channels, num_channels, kernel_size=(1, 1)))
         elif num_bands == 4:
             print('Finetuning pretrained deeplabv3 with 4 bands')
-            model = models.segmentation.deeplabv3_resnet101(pretrained=pretrained, progress=True, aux_loss=None)
+            model = models.segmentation.deeplabv3_resnet101(pretrained=True, progress=True)
```
CharlesAuthier, line 66: Same comment than before.

models/model_choice.py
```diff
@@ -57,17 +57,21 @@ def net(net_params, num_channels, inference=False):
         assert (num_bands == 3 or num_bands == 4), msg
         if num_bands == 3:
             print('Finetuning pretrained deeplabv3 with 3 bands')
-            model = models.segmentation.deeplabv3_resnet101(pretrained=pretrained, progress=True, aux_loss=None)
-            model.classifier = common.DeepLabHead(2048, num_channels)
+            model = models.segmentation.deeplabv3_resnet101(pretrained=True, progress=True)
+            classifier = list(model.classifier.children())
+            model.classifier = nn.Sequential(*classifier[:-1])
```
CharlesAuthier, line 62: Just be sure that the `-1` is not the `tanh()` layer and really the `conv2d`.

utils/augmentation.py
remtav: what do you mean? Radiometric Trim augmentation should not have any impact on scaling: the output values of RadiometricTrim are in the same range as the input values (ex.: 0 to 255 for uint8), so scaling function doen's need to "know" if radiomtric trim is being performed.

utils/utils.py
remtav: see previous answer

utils/verifications.py
remtav: is_valid_geom was already part of rasterio, but had a bug. So I had copied the function to GDL and fixed the bug temporarily while rasterio took care of that. Now the bug is solved. Rasterio has been update on cavum/velum and HPC, so we shouldn't encounter this bug again.

models/model_choice.py
```diff
@@ -57,17 +57,21 @@ def net(net_params, num_channels, inference=False):
...
+            model.classifier.add_module('4', nn.Conv2d(classifier[-1].in_channels, num_channels, kernel_size=(1, 1)))
```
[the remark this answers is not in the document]
CharlesAuthier: Keep it like that for now, we may change it in a futur version.

images_to_samples.py
mpelchat04: There are occasions where the raster's metadata are empty. In my case it was an error with my Gdal installation and Rasterio wasn't throwing any errors with empty CRS in the raster.   ([fix for GDAL here](https://stackoverflow.com/questions/56764046/gdal-ogr2ogr-cannot-find-proj-db-error#:~:text=3%20Answers&text=You%20might%20need%20to%20set%20the%20PROJ_LIB%20environment%20variable.&text=If%20you%20don't%20find,see%20if%20things%20start%20working))

**Question is :** Are there negative impacts in our process if we have incorrect metadata?   
- If so, we should handle such case.   
- If we don't do much with the metadata (it's just info about the original raster), a simple warning could do.

remtav: the metadata is only used for 2 things during training: 
1. knowing what's the original datatype for scaling (ex.: if scaling knows original range is 0-255, it will then scale all values from that to 0-1. An exception handling behavior if metadata is missing would be to scan the min and max values in array and deduce a datatype based on that. Would that be acceptable ?
2. random radiometric trim: the histogram of source image is used to determine which value is at 2nd percentile, for example. It keeps GDL from computing histogram from cropped sample, which is not representative enough and is also longer to compute. There can't be any problem with this step though, since the histogram can always be computed at sample prep time: it's just a bincount on the source image saved to a dictionary.

mpelchat04: I don't think it's required at this time. Maybe we can file an issue about it and we'll discuss the best way to go from there.  
GDAL will raise a warnings for the proj.db file missing, related to the issue I got.

victorlazio109: I would run some tests on this.

utils/augmentation.py
victorlazio109: okay, sounds good.

Issue #110, opened 2019-11-04
Inference: cannot infer on large images (>3Gb)


remtav:
When using inference.py with large model (ex.: deeplabv3) and image (>3Gb), process saturates RAM on cpu and gets kill by OS. 

Idea to fix bug: inference should write to disk (ex.: empty .tif) rather than keep process in RAM.

Issue #52, opened 2019-02-06
"sqlite3_open" error whith geopackage


mpelchat04:
The error occur in [images_to_samples.py](https://github.com/NRCan/geo-deep-learning/blob/master/images_to_samples.py#L209), when we have more than 1000 images to prepare.  
```shell
...  
Images/mtq_f09_257000_5077000_50_int_rou_asp_slo.tif
{'trn': 42601, 'val': 12404}
Images/15_2525134f09_dc_50_int_rou_asp_slo.tif
{'trn': 42601, 'val': 12501}ERROR 4: sqlite3_open(chapeau_mtm09_lac_riv_build.gpkg) failed: unable to open database file

Images/mtq_f09_260000_5081000_50_int_rou_asp_slo.tif
{'trn': 42601, 'val': 12516}
Images/15_2495160f09_dc_45_50_int_rou_asp_slo.tif
{'trn': 42601, 'val': 12714}
Images/15_2575131f09_dc_50_int_rou_asp_slo.tif
{'trn': 42601, 'val': 12714}
Images/15_2605131f09_dc_50_int_rou_asp_slo.tif
{'trn': 42601, 'val': 12756}
Images/15_2455150f09_dc_50_int_rou_asp_slo.tif
{'trn': 42601, 'val': 12793}
Images/mtq_f09_262000_5093000_50_int_rou_asp_slo.tif
{'trn': 42601, 'val': 12799}
Images/15_2715113f09_dc_50_int_rou_asp_slo.tif
{'trn': 42601, 'val': 12854}
Images/15_2505138f09_dc_50_int_rou_asp_slo.tif
{'trn': 42601, 'val': 12858}
Images/15_2595126f09_dc_50_int_rou_asp_slo.tif
{'trn': 42601, 'val': 12888}
Traceback (most recent call last):
  File "images_to_samples.py", line 258, in <module>
    params['sample']['mask_reference'])
  File "images_to_samples.py", line 209, in main
    vector_to_raster(info['gpkg'], info['attribute_name'], label_raster)
  File "images_to_samples.py", line 133, in vector_to_raster
    source_layer = source_ds.GetLayer()
AttributeError: 'NoneType' object has no attribute 'GetLayer'
```  
One of the possible solutions might include:  
- Using [Fiona](https://fiona.readthedocs.io/en/latest/README.html) to read/write geopackages : Fiona is OGR’s neat and nimble API for Python programmers.  
[...]

PR #149, opened 2020-07-28
Benchmarking and inference in GDL

victorlazio109:
Integration of stand-alone benchmarking platform into GDL while optimizing and extending it's functionalities.


.gitignore
```diff
@@ -1,2 +1,3 @@
 *__pycache__**
```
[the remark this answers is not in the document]
victorlazio109: we currently have pycache in the develop branch. which is generated at runtime. I thought this are auxiliary files we would not want on the repo?

inference.py
```diff
@@ -266,12 +176,20 @@ def main(params: dict):
...
+    # CONFIGURE MODEL
+    model, state_dict_path, model_name = net(params, num_channels=num_classes_corrected, inference=True)
     try:
         model.to(device)
     except RuntimeError:
         print(f"Unable to use device. Trying device 0")
         device = torch.device(f'cuda:0' if torch.cuda.is_available() and lst_device_ids else 'cpu')
         model.to(device)
+    
+    # mlflow tracking path + parameters logging
```
mpelchat04, line 188: Yeah!

inference.py
```diff
@@ -221,42 +152,21 @@ def main(params: dict):
...
-
-    debug = get_key_def('debug_mode', params['global'], False)
-    if debug:
-        warnings.warn(f'Debug mode activated. Some debug features may mobilize extra disk space and cause delays in execution.')
-
-    num_classes = params['global']['num_classes']
     task = params['global']['task']
-    num_classes_corrected = add_background_to_num_class(task, num_classes)
-
+    img_dir_or_csv = params['inference']['img_dir_or_csv_file']
```
mpelchat04, line 156: If we only provide a path to a folder, does that means that only tifs will be read (i.e. no gpkg of potential benchmark dataset)?  

If so, we should include this behavior in the README, in the inference section. :)

inference.py
```diff
@@ -16,133 +16,66 @@
...
+                                                    input_image, 
+                                                    meta_map=None, 
+                                                    raster_info=None)
+    h, w, num_bands = img_array.shape
+    h_ = sample_size * math.ceil(h / sample_size)
+    w_ = sample_size * math.ceil(w / sample_size)
+    pred_img = np.empty((h_, w_), dtype=np.uint8)
+    for row in tqdm(range(0, h, sample_size), position=1, leave=False, desc='Inferring rows'):
+        with tqdm(range(0, w, sample_size), position=2, leave=False, desc='Inferring columns') as _tqdm:
+            for column in _tqdm:
```
mpelchat04, line 53: Since you're not using the overlap parameter, have you noticed inconsistencies in the resulting raster, like those showed in https://github.com/NRCan/geo-deep-learning/issues/11

inference.py
```diff
@@ -221,42 +152,21 @@ def main(params: dict):
...
-        warnings.warn(f'Debug mode activated. Some debug features may mobilize extra disk space and cause delays in execution.')
-
-    num_classes = params['global']['num_classes']
     task = params['global']['task']
-    num_classes_corrected = add_background_to_num_class(task, num_classes)
-
+    img_dir_or_csv = params['inference']['img_dir_or_csv_file']
     chunk_size = get_key_def('chunk_size', params['inference'], 512)
-    overlap = get_key_def('overlap', params['inference'], 10)
-    nbr_pix_overlap = int(math.floor(overlap / 100 * chunk_size))
```
mpelchat04, line 235: The overlap parameter isn't used anymore, right? 

If so, we should remove it from the README and the template yaml config file.

utils/metrics.py
```diff
@@ -93,3 +93,100 @@ def iou(pred, label, batch_size, num_classes, metric_dict, only_present=True):
     mean_IOU = np.nanmean(ious)
     metric_dict['iou'].update(mean_IOU, batch_size)
     return metric_dict
+
+#### Benchmark Metrics ####
```
mpelchat04, line 97: Clean! Well done!

inference.py
victorlazio109: Yes, that is the current implementation. I would document this behaviour in the README.md. tnx

inference.py
victorlazio109: Hmmn, I had no idea the overlap solves that problem. I wanted to implement this https://github.com/Vooban/Smoothly-Blend-Image-Patches to solve that problem. If the overlap solves that problem I would be happy to retain it.

inference.py
victorlazio109: see comment on line #52

inference.py
mpelchat04: Well, the overlap reduced the impact of this problem but did not solved it completely.  

Reading what's in the link you provided, I really prefer your approach. The first advantage I see is that it allows for data parallelization, which @remtav will certainly be glad to see!

mpelchat04:
Have you tested with images larger than 3Gb? Do we still have this issue #110 ?

utils/readers.py
```diff
@@ -47,13 +47,10 @@ def image_reader_as_array(input_image,
...
         np_array, input_image = clip_raster_with_gpkg(input_image, clip_gpkg, debug=debug)
-        np_array = np.transpose(np_array, (1, 2, 0))  # send channels last
-
     else:
-        np_array = np.empty([input_image.height, input_image.width, input_image.count], dtype=np.uint16)
-        for i in tqdm(range(input_image.count), position=1, leave=False, desc=f'Reading image bands: {Path(input_image.files[0]).stem}'):
-            np_array[:, :, i] = input_image.read(i+1)  # Bands starts at 1 in rasterio not 0
+        np_array = input_image.read()
 
+    np_array = np.moveaxis(np_array, 0, -1) # send channels last
```
remtav, line 53: You are sure the .read() function returns an array with channels in first dimension?

utils/utils.py
```diff
@@ -155,21 +155,32 @@ def get_key_def(key, config, default=None, msg=None, delete=False, expected_type
...
-    :param scale_range: Desired range of transformed data.
+    :param scale_range: Desired range of transformed data (0, 1) or (-1, 1).
     :param orig_range: Original range of input data.
     :return: (numpy array) Scaled image
     """
-    # range(0, 1)
-    scale_img = (img - orig_range[0]) / (orig_range[1] - orig_range[0])
-    # range(min_value, max_value)
-    scale_img = scale_img * (scale_range[1] - scale_range[0]) + scale_range[0]
+    assert scale_range == (0, 1) or scale_range == (-1, 1), 'expects scale_range as (0, 1) or (-1, 1)'
```
remtav, line 166: I like the idea of giving only two possible output scales.

remtav:
Very good Vic. It didn't need as mush refactoring as I though it would in the end.

utils/readers.py
victorlazio109: yes, gdal and rasterio behaviour

utils/metrics.py
```diff
@@ -93,3 +93,100 @@ def iou(pred, label, batch_size, num_classes, metric_dict, only_present=True):
...
+        return dsc
+    
+    @staticmethod
+    def precision(label, pred):
+        '''
+        :return: precision
+        '''
+        tp = int(np.sum(np.logical_and(label, pred)))
+        fp = int(np.sum(np.logical_and(pred, np.logical_not(label))))
+        p = tp / (tp + fp)
```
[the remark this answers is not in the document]
victorlazio109: all seems fine now, please raise any of these issues/bugs you encounter.

utils/metrics.py
```diff
@@ -93,3 +93,100 @@ def iou(pred, label, batch_size, num_classes, metric_dict, only_present=True):
...
+    @staticmethod
+    def jaccard(label, pred):
+        '''
+        :return: IOU
+        '''
+        both = np.logical_and(label, pred)
+        either = np.logical_or(label, pred)
+        ji = (np.sum(both) + min_val) / (np.sum(either) + min_val)
+        
+        return ji
```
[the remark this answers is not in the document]
victorlazio109: No, at least not at this moment. could be refactored in the near future.

data/inference_sem_seg_ci_csv.csv
```diff
@@ -1,2 +1,2 @@
-./data/22978945_15.tif,
-./data/23429155_15.tif,
+./data/22978945_15.tif,,,
+./data/23429155_15.tif,,./data/massachusetts_buildings.gpkg,properties/class
```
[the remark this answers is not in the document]
victorlazio109: the gpkg is added because of the benchmarking functionality included in inference.py. inference.py now handles individual images with or without gpkg for benchmarking. The README.md has been updated which should help.

conf/config_template.yaml
```diff
@@ -97,8 +97,8 @@ inference:
   img_dir_or_csv_file: /path/to/csv/containing/images/list.csv
   state_dict_path: /path/to/model/weights/for/inference/checkpoint.pth.tar
   chunk_size: 512 # (int) Size (height and width) of each prediction patch. Default: 512
-  overlap: 10 # (int) Percentage of overlap between 2 chunks. Default: 10
-  heatmaps: True
+  smooth_prediction: True
+  overlap: 2 # overlap between tiles for smoothing. Must be an even number that divides chunk_size without remainder.
```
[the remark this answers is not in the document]
victorlazio109: inference overlap is in number of pixels not %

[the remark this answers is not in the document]
remtav: Can we have it as a % instead? It would become uniform with [images_to_samples](https://github.com/NRCan/geo-deep-learning/blob/develop/images_to_samples.py#L214)
For images_to_samples.py, it used to be number of pixels, but we realized it was more intuitive to have it as a percentage.

inference.py
```diff
@@ -16,133 +16,117 @@
...
-        index: (int) index of array from list of images on which inference is performed
-
-        returns a numpy array of the same size (h,w) as the input image, where each value is the predicted output.
-    """
-
+colors = {0: [0, 0, 0],  # Background
+          1: [27, 120, 55],  # Vegetation
+          2: [116, 173, 209], # Hydro
+          3: [223, 194, 125],  # Roads
+          4: [150, 150, 150]}  # Buildings
```
remtav, line 39: why hardcode this? won't it cause bug it num_classes > 5 ?
visualization.py was implemented to always [default to a colormap defined by matplotlib](https://github.com/NRCan/geo-deep-learning/blob/develop/utils/visualization.py#L138).

inference.py
```diff
@@ -16,133 +16,117 @@
...
+    metadata = add_metadata_from_raster_to_sample(img_array, 
+                                                    input_image, 
+                                                    meta_map=None, 
+                                                    raster_info=None)
+    h, w, bands = img_array.shape
+    assert num_bands <= bands, f"Num of specified bands is not compatible with image shape {img_array.shape}"
+    if num_bands < bands:
+       img_array = img_array[:, :, :num_bands]
+    
+    padding = int(round(sample_size * (1 - 1.0/overlap)))
```
remtav, line 56: hmm, I don't understand this. If your overlap is set to "2" as default, you're actually padding with 50% of the sample size. Are you sure you didn't mean:
dist_samples = int(round(sample_size * (1 - (overlap / 1.0)))
If so, then your considering the overlap as a ratio, not a number of pixels. Still, the default value of "2" wouldn't make sense.
I would suggest keeping it like this, but having it as a percentage:
dist_samples = round(sample_size * (1 - (overlap / 100)))
(the "int" conversion isn't necessary, since "round" already outputs an int...)
the default value could be 25 (ie 25%), for example.

inference.py
```diff
@@ -305,80 +276,57 @@ def main(params: dict):
...
-
+                    print(inference_image)                    
+                assert local_img.is_file(), f"Could not locate raster file at {local_img}"
+                with rasterio.open(local_img, 'r') as raster:
+                    inf_meta= raster.meta
+                    if prediction_with_smoothing:
+                        print('Smoothening Predictions with 2D interpolation')
+                        pred = segmentation_with_smoothing(raster, local_gpkg, model, chunk_size, overlap, num_bands, device)
+                    else:
+                        pred = segmentation(raster, local_gpkg, model, chunk_size, num_bands, device)
```
remtav, line 287: without any overlap, I'm afraid inference could be disappointing...

inference.py
remtav: Parallelization would be cool. However I don't see the interest in using segmentation without smoothing. We're keeping only as a basice version of inference?

victorlazio109: Yes, plus the smoothing can be fine-tuned better. Nice to have a base version for comparison. it can be deprecated in the future.

inference.py
victorlazio109: it's hardcoded right now because the color mapping function is only efficient for smaller arrays. I haven't found a way to optimize it using matplotlib.

inference.py
victorlazio109: How the padding is calculated in this case is slightly different, the intention would be to use a 'valid' border padding strategy instead of 'same' padding with zeros. In this case, np.pad(mode='reflect') can be used.  https://numpy.org/doc/stable/reference/generated/numpy.pad.html. Because of this, there are hard limits set on the overlap: In the README.md:  'overlap between tiles for smoothing. Must be an even number that divides chunk_size without remainder'. Also round does not always output int, take a look at this official doc on round in Python3 https://docs.python.org/3/library/functions.html#round

I understand the confusion with the overlap param and how it clashes with the one set in 'sample'. Perhaps I could change the name of this param to avoid such.

inference.py
victorlazio109: Yeah, two versions of prediction functions are offered now. One with overlaps and smoothening and another base version without smoothening.

inference.py
CharlesAuthier: Maybe use a colormap and pick the number of color associate to the number of classes. 
```
import matplotlib.pyplot as plt
import random

number_of_colors = 8

color = ["#"+''.join([random.choice('0123456789ABCDEF') for j in range(6)])
             for i in range(number_of_colors)]
print(color)
```
(random exemple)

utils/metrics.py
```diff
@@ -1,7 +1,8 @@
-from sklearn.metrics import classification_report
 import numpy as np
+from sklearn.metrics import classification_report, matthews_corrcoef, recall_score
+from math import sqrt
```
CharlesAuthier, line 3: Instead of using the math package, use the `np.sqrt`, it will avoid unnecessary import.

inference.py
victorlazio109: The problem lies in the mapping function, not the color picker. I don't have an efficient way of mapping HTML colors without Matplotlib. It is simply not efficient for large arrays.

CharlesAuthier: Ok I understand :/ hummm

remtav: Ok. I guess we can hardcode as a temporary approach, but it would be important to handle tasks with more than 4 classes. I'll let you figure what solution fits best.

inference.py
remtav: Ok. My only (final!) concern is that infering with an overlap (commonly 25%) in the previous version of inference.py gave much better results than no overlap (i.e. what the base version would output). 
Although it's great in terms of memory efficiency and necessary computing, I'm not convinced the smoothing version will give an output just as good as inference with 25% overlap. On the demo you gave, I saw quite a few edges where smoothing introduced other values (ex.: vegetation in hydro or the opposite). I'd have to see it again, but anyways my concern is smoothing may not perform as well as previous inference with overlap.

victorlazio109: I think the base version can be included with overlap. No problems. I can create a new ticket for that.

PR #162, opened 2020-09-09
K8s dev

krishanr:
This pull request adds a Dockerfile as well as instructions for building the Dockerfile and running the scripts images_to_samples.py, train_segmentation.py, and inference.py in the docker image. I've also added instructions for building and running the image on remote computers.


victorlazio109:
Do you mind bringing your PR up to date as I would like to merge the latest updates from 'develop' into 'K8s' branch I don't think there would be a lot of conflicts with your PR after the update. Based on this I think travis-CI test would pass effortlessly.

ymoisan:
Looking at your [Dockerfile](https://github.com/NRCan/geo-deep-learning/blob/a3ef3c7e7e86add72af8ed4d9f6c3be1b419e11e/Dockerfile#L2), I pulled the image you reference on my machine.  It takes up 12 GB of space even though it's advertised at 4.5GB on the [nvidia site](https://ngc.nvidia.com/catalog/containers/nvidia:pytorch/tags).  I'm worried about the time spinning such big containers for a real online inference service.  

The way we run GDL on our workstations and HPC is we set up cuDNN on the servers themselves, that is system wide.  I understand talking with @victorlazio109 that it may be tricky to get the right version of cuDNN depending on the version of CUDA used by a specific version of PyTorch (am I right @victorlazio109 ?).  

I think we should strive to get a lean Dockerfile for GDL (that is, basically wrap our current conda environment in a container) and take for granted our GPU servers are properly set to run the version of PyTorch we want to use in GDL.  Since we will be using a container orchestrator (K8S via Rancher) to run our inference workloads, I think the responsibility of finding the right GPU node should reside in Rancher.  I'm not sure how we can add the right metadata to our K8S cluster nodes so that Rancher knows a node has the right version of cuDNN for the container it is asked to deploy, but I think we should look in that direction.  Even though it may be quicker than I think to launch 12 GB containers, it does not make sense to me to rebuild a GPU runtime that's already on a server.

To that effect, we will be investigating the use of Kubeflow through Rancher soon.  I'll keep you informed when we have something working.  In the meantime, if you could investigate more "efficient" ways to put together a Docker image, that would be great.

PR #165, opened 2020-10-13
add spacenet models

CharlesAuthier:


models/model_choice.py
```diff
@@ -141,6 +141,28 @@ def net(net_params, num_channels, inference=False):
             in_channels=num_bands,
             classes=num_channels,
             activation=None)
+    elif model_name == 'spacenet_unet_efficientnetB5_pretrained':
```
victorlazio109, line 144: I think we can create a markdown file to document potential models we can use with smp. It would be counterproductive to define so many models as code here since they can easily be called by defining smp parameters.

CharlesAuthier: Good,  I had the intention to make some MD in some folder for a better documentation, I dont know when I will have the time but yes.

mpelchat04 (dismissed):

remtav (approved):

mpelchat04 (approved):

victorlazio109 (commented):
avoids duplication, epic!

victorlazio109 (approved):

PR #169, opened 2020-12-28
added allow duplicates in yaml file

mkutu:
added yaml.allow_duplicates_keys in /conf/readers.py (line 18) to allow duplicate "overlap" inference parameters in the config_template.yaml


images_to_samples.py
```diff
@@ -33,7 +33,7 @@ def mask_image(arrayA, arrayB):
     >>> x2 = np.array([1.5, 1.2, 1.6, 1.2, 11., 1.1, 25.9, 0.1], dtype=np.float32).reshape(2,2,2)
     >>> mask_image(x1, x2)
     array([[[ 0. ,  0. ],
-        [ 1.6,  1.2]],
+        [.6,  1.2]],
```
CharlesAuthier, line 36: Why changing that?

images_to_samples.py
```diff
@@ -310,21 +310,23 @@ def main(params):
...
-            final_samples_folder = data_path.joinpath("samples")
-        else:
-            final_samples_folder = "samples"
-        samples_folder = sample_path_name
-
-    else:
-        list_data_prep = read_csv(csv_file)
-        samples_folder = data_path.joinpath(sample_path_name)
-
+    # if bucket_name:
```
CharlesAuthier, line 313: Why commenting? If you don't put a bucket name it's not suppose to execute.

images_to_samples.py
```diff
@@ -383,18 +385,18 @@ def main(params):
...
-                    if info['gpkg'] not in bucket_file_cache:
-                        bucket_file_cache.append(info['gpkg'])
-                        bucket.download_file(info['gpkg'], info['gpkg'].split('/')[-1])
-                    info['gpkg'] = info['gpkg'].split('/')[-1]
-                    if info['meta']:
-                        if info['meta'] not in bucket_file_cache:
-                            bucket_file_cache.append(info['meta'])
-                            bucket.download_file(info['meta'], info['meta'].split('/')[-1])
-                        info['meta'] = info['meta'].split('/')[-1]
+                # if bucket_name:
```
CharlesAuthier, line 388: Same comment than before.

images_to_samples.py
```diff
@@ -503,11 +505,11 @@ def main(params):
 
     print("Number of samples created: ", number_samples)
 
-    if bucket_name and final_samples_folder:
-        print('Transfering Samples to the bucket')
-        bucket.upload_file(samples_folder + "/trn_samples.hdf5", final_samples_folder + '/trn_samples.hdf5')
-        bucket.upload_file(samples_folder + "/val_samples.hdf5", final_samples_folder + '/val_samples.hdf5')
-        bucket.upload_file(samples_folder + "/tst_samples.hdf5", final_samples_folder + '/tst_samples.hdf5')
+    # if bucket_name and final_samples_folder:
```
CharlesAuthier, line 508: Same comment than before.

models/model_choice.py
```diff
@@ -41,6 +41,7 @@ def net(net_params, num_channels, inference=False):
     pretrained = get_key_def('pretrained', net_params['training'], True) if not inference else False
     dropout = get_key_def('dropout', net_params['training'], False)
     dropout_prob = get_key_def('dropout_prob', net_params['training'], 0.5)
+    num_classes = net_params['global']['num_classes']
```
CharlesAuthier, line 44: The `num_classes` is the ` num_channels`.

models/model_choice.py
```diff
@@ -6,7 +6,7 @@
 import torch.nn as nn
 import segmentation_models_pytorch as smp
 import torchvision.models as models
-from models import TernausNet, unet, checkpointed_unet, inception, coordconv
+from models import TernausNet, unet, checkpointed_unet, inception, coordconv, fcn
```
CharlesAuthier, line 9: If you have created a file `.py` for the fc you can add it to the pull request.

CharlesAuthier (changes requested):

CharlesAuthier:
- Please don't push `conf/config_template.yaml` instead create a config file that is a copy of  `conf/config_template.yaml` with your parameters. Run your config file for testing, but don't add the file to the pull request please.
- Why do you switch the colors? If it's a personal choice, just add the file in the `.gitgnore` during the pull request.

Otherwise, I don't see why not allow the `allow_duplicate_keys = True`

utils/CreateDataset.py
```diff
@@ -73,6 +73,8 @@ def __init__(self, work_folder,
                 metadata = hdf5_file["metadata"][i, ...]
                 if isinstance(metadata, np.ndarray) and len(metadata) == 1:
                     metadata = metadata[0]
+                    metadata = metadata.decode('ascii')
```
CharlesAuthier, line 76: @victorlazio109 @remtav can you look at this i am not familiar with this file.

CharlesAuthier:
Perfect, thanks for the changes, the only thing left to do before merging is to create a `conf/travis_CI/config_ci_segmentation_local_fcn.yaml`, you can copy `conf/travis_CI/config_ci_segmentation_local.yaml` and change the model when it will pass the Travis test we will merge. Don't hesitate if you have any question.

utils/readers.py
```diff
@@ -15,6 +15,7 @@ def read_parameters(param_file):
         YAML (Ruamel) CommentedMap dict-like object
     """
     yaml = YAML()
+    yaml.allow_duplicate_keys=True
```
remtav, line 18: Can you give an example when this is necessary? Wouldn't we want to raise an error if there are duplicate keys?

utils/CreateDataset.py
remtav: @mpelchat04 and @felegare have worked more than @victorlazio109 and I with metdata. Any input?
@mkutu I'm guessing this line was necessary for your use of metadata?

requirements.txt
```diff
@@ -15,3 +15,5 @@ tqdm
 matplotlib
 geopandas
 fiona
+segmentation_models_pytorch
+mlflow
```
remtav, line 19: indeed, these should be added.

models/model_choice.py
```diff
@@ -63,6 +63,9 @@ def net(net_params, num_channels, inference=False):
         model = models.segmentation.fcn_resnet101(
             pretrained=pretrained, progress=True, num_classes=num_channels, aux_loss=None
         )
+    elif model_name == 'fcn_vgg':
+        vgg_model = fcn_vgg.VGGNet(requires_grad=True)
+        model = fcn_vgg.FCNs(pretrained_net=vgg_model, n_class=num_channels)
```
remtav, line 68: Nice addition. Did you experience better results with FCN/VGG than with other common semantic segmentation models (Deeplabv3, deeplav3+, Unet/Resnet, etc.)?

utils/readers.py
mkutu: yea,  conf/config_template.yaml has two key instances of "overlap". I've added the bit below. Perhaps it would be easier to relabel them. 

https://github.com/NRCan/geo-deep-learning/blob/6852729d0ec03c0c2a6465b7a68039181c8bcbda/conf/config_template.yaml#L113-L123

models/model_choice.py
mkutu: Nothing concrete, but I've done some prelim training on leaf-level scale where I've seen improved performance over Unet/resnet. Haven't tried deeplabv3s yet. My training ability is a little limited right now.

utils/CreateDataset.py
mkutu: yes, this was necessary for me. Perhaps I should make another issue for this? 
I was getting, AttributeError: 'bytes' object has no attribute 'update'

I could not begin training because the hdf5 file was being read as non-unicode string which python displays as bytes ('b').

utils/readers.py
remtav: Thanks for bringing that to our attention. There's shouldn't be duplicate keys here and I'm afraid allowing them would not provide the desired result as, according to [the yaml documentation](https://yaml.readthedocs.io/en/latest/api.html), the second instance of the same key would be discarded. I'll create a "bug" issue I guess so we can correct this ASAP.

Issue #177, opened 2021-01-12
Repeatability: in config yaml, add git tag associated with the version of GDL being used for train


remtav:
To help with repeatability, config should contain all necessary information to repeat a training and obtain exactly the same model (if randomness in gdl is controlled, which it is not currently. Should it? Is it possible/desired with DL models?)

Minimally, yaml should contain info about which version of GDL was used to train. How to make sure the mentionned version is actually the one used? Food for thought...

Thanks @bstdenis for raising.

bstdenis:
If you properly set all the random seed at the appropriate places in the code, a training procedure can be repeated exactly on the same machine, but there is no guarantee on another machine. In my opinion, the main issue is more with having a piece of code that actually works for an old experiment. As the system evolves (change version) it is possible that old experiments will become broken with the new code, so keeping a link (tag, github hash, etc.) of the version is good practice. If it's an experiment that uses work-in-progress code, it's still a good idea to at least push it in a development branch and we can then link it with its commit hash value.

ymoisan:
Folks, this should be taken care of in the metadata of the model, to be stored eventually in a STAC extension.  The CRIM has begun laying out a proposal for a new STAC extension for Deep Learning Models (DLM).  The hash code of a particular model instanciation is [here](https://github.com/crim-ca/CCCOT03/blob/cebb8153a041bb08b3663e15eaf6467c1d2b954b/extension/dl-model/json-schema/schema.json#L90).  We expect to propose the extension to STAC more formally in the coming weeks.

PR #181, opened 2021-02-25
New parameters for mlflow and directory structure

bstdenis:
Add parameters to customize mlflow information, and use it to better differentiation images_to_samples directories
Add parameters to customize minimum GPU availability
Fix doctest use in mask_image
Fix edge case of all nan IOUs in a tile


images_to_samples.py
```diff
@@ -33,9 +33,10 @@ def mask_image(arrayA, arrayB):
     >>> x2 = np.array([1.5, 1.2, 1.6, 1.2, 11., 1.1, 25.9, 0.1], dtype=np.float32).reshape(2,2,2)
     >>> mask_image(x1, x2)
     array([[[ 0. ,  0. ],
-        [ 1.6,  1.2]],
-        [[11. ,  1.1],
-        [25.9,  0.1]]], dtype=float32)
+            [ 1.6,  1.2]],
```
bstdenis, line 36: The result in the docstring is not coherent with what the code does. Can anyone confirm that the actual output (which I fixed here) is what we want from this function?

train_segmentation.py
```diff
@@ -458,7 +460,8 @@ def main(params, config_path):
     overlap = params["sample"]["overlap"]
     min_annot_perc = get_key_def('min_annotated_percent', params['sample']['sampling_method'], 0, expected_type=int)
     num_bands = params['global']['number_of_bands']
-    samples_folder_name = f'samples{samples_size}_overlap{overlap}_min-annot{min_annot_perc}_{num_bands}bands'  # FIXME: won't check if folder has datetime suffix (if multiple folders)
```
bstdenis, line 461: I propose to use a config parameter to avoid confusion with the samples folder names that share parameters but not image sources.

conf/config_template.yaml
```diff
@@ -16,6 +16,8 @@ global:
   # Model available: unet, unetsmall, checkpointed_unet, ternausnet, fcn_resnet101, deeplabv3_resnet50, deeplabv3_resnet101
   model_name: unet
   mlflow_uri: path/to/mlflow_tracking_uri  # Default to ./mlruns
+  mlflow_experiment_name: gdl-training
+  mlflow_run_name: gdl  # also used for naming of preprocessing directories
```
remtav, line 20: super relevant, thanks!

conf/config_template.yaml
```diff
@@ -30,6 +32,8 @@ global:
   debug_mode: True
   coordconv_convert: False
   coordvonc_scale:
+  max_used_ram: 2000
+  max_used_perc: 15
```
remtav, line 36: would add a comment (in elsewhere in doc) to tell user what default values are (and of course that these values as useless if num_gpu=0)

conf/config_template.yaml
```diff
@@ -30,6 +32,8 @@ global:
...
 
 
 # Data analysis parameters; used in data_analysis.py ------------------
```
remtav, line 39: seeing that you're modifying config_template.yaml, do you want to address issue #189 in this PR?

images_to_samples.py
remtav: I didn't test this docstring... who added this? @mpelchat04 do you know? This function was implemented before I arrived at CCCOT

images_to_samples.py
```diff
@@ -326,8 +328,7 @@ def main(params):
         samples_folder = data_path.joinpath(sample_path_name)
 
     if samples_folder.is_dir():
-        warnings.warn(f'Data path exists: {samples_folder}. Suffix will be added to directory name.')
-        samples_folder = Path(str(samples_folder) + '_' + now)
+        raise FileExistsError(f'Data path exists: {samples_folder}. Remove it or use a different run_name.')
```
remtav, line 331: I agree to change this behavior. During training, I think it's ok to automatically add a suffix if the directory already exists. During sample preparation, the opposite behavior (raise exception) makes sense. An alternative approach would be to keep old behavior only if in "debug" mode. I found it useful to have an automatic directory creation when I was debugging in the past...

conf/config_template.yaml
remtav: and how about address #177, i.e. add a "gdl_git_hash" parameter to "save" git version for repeatability ?

train_segmentation.py
remtav: I totally agree. How would you name the samples folder then?

train_segmentation.py
```diff
@@ -491,7 +494,12 @@ def main(params, config_path):
     num_devices = params['global']['num_gpus']
     assert num_devices is not None and num_devices >= 0, "missing mandatory num gpus parameter"
     # list of GPU devices that are available and unused. If no GPUs, returns empty list
-    lst_device_ids = get_device_ids(num_devices) if torch.cuda.is_available() else []
+    max_used_ram = get_key_def('max_used_ram', params['global'], 2000)
+    max_used_perc = get_key_def('max_used_perc', params['global'], 15)
```
remtav, line 498: nice use of get_key_def(). You could even force an "int" as input with the [expected_type](https://github.com/NRCan/geo-deep-learning/blob/8d85078a1aeae791244807660c5cd5119cca7d83/utils/utils.py#L126) argument. Just a suggestion.

remtav (commented):
nice additions, thanks!

train_segmentation.py
bstdenis: currently I implemented it by using the new mlflow_run_name parameter (when given), not entirely sure this is the best way

conf/config_template.yaml
bstdenis: yes, propositions made in the latest commits

train_segmentation.py
```diff
@@ -540,7 +548,11 @@ def main(params, config_path):
 
     criterion = criterion.to(device)
 
-    filename = output_path.joinpath('checkpoint.pth.tar')
+    gdl_git_ref = get_key_def('gdl_git_reference', params['global'], None)
+    if gdl_git_ref is not None:
+        filename = output_path.joinpath(f'checkpoint_{Path(gdl_git_ref).name}.pth.tar')
```
bstdenis, line 553: I propose to add the git tag or commit hash to the checkpoint name (taken from new gdl_git_reference config parameter when available). This would require also paying attention to the state_dict_path config at inference time.

images_to_samples.py
remtav: second though, for debugging, it's really fun to not raise an exception if directory exists (i.e. when having to run images_to_samples multiple times in a small time lapse)

train_segmentation.py
remtav: that's a good approach. I just suggested another approach in my PR #191. Take a look and let me know. In my PR, I don't ask user for git hash, I just getting from git command and store in hdf5 and checkpoint.tar.pth (after adding it to params)

images_to_samples.py
mpelchat04: Yup, my bad.

config/config_template.yaml
```diff
@@ -32,6 +36,7 @@ global:
   debug_mode: True
   coordconv_convert: False
   coordvonc_scale:
+  gdl_git_reference: https://github.com/NRCan/geo-deep-learning/tree/v.1.2.0  # git tag or commit hash of the version of gdl used
```
mpelchat04, line 39: Good call.

mpelchat04 (approved):

images_to_samples.py
bstdenis: my confusion is that train_segmentation always uses the first one (without the added datetime), so it feels like rerunning images_to_samples is a dead end, because if you like the new results, they are not in a directory used by train_segmentation. Perhaps debug mode should instead move the old directory (with a random suffix?). To me it seems like a mistake waiting to happen to rerun images_to_samples, and then run train_segmentation thinking the new samples will be used.

train_segmentation.py
bstdenis: Your approach is probably better, the fewer things we have to enter manually in the config, the less likely we are to just leave it unchanged when making a new config. Let's get your solution through and I will revert my stuff accordingly.

config/config_template.yaml
```diff
@@ -18,9 +18,13 @@ global:
...
   mlflow_uri: ./mlruns  # path/to/mlflow_tracking_uri
-  bucket_name:   # name of the S3 bucket where data is stored. Leave blank if using local files
+  mlflow_experiment_name: gdl-training
+  mlflow_run_name: gdl  # also used for naming of preprocessing directories
+  bucket_name:  # name of the S3 bucket where data is stored. Leave blank if using local files
   task: segmentation  # Task to perform. Either segmentation or classification
-  num_gpus: 0 # If 0, will run on CPU
+  num_gpus: 0  # If 0, will run on CPU
+  max_used_ram: 2000  # maximum used GPU memory (MB) in order to use it (only if num_gpus > 0)
+  max_used_perc: 15  # maximum utilization rate (percent) of the GPU in order to use it (only if num_gpus > 0)
```
CharlesAuthier, line 27: i like it

config/config_template.yaml
```diff
@@ -63,8 +68,8 @@ training:
   num_trn_samples:
   num_val_samples:
   num_tst_samples:
-  batch_size: 2
-  num_epochs: 1
+  batch_size: 4
```
CharlesAuthier, line 71: maybe keep the `batch_size` at 2 and the `num_epochs` at 1 just for people who want to test out of the box for the tamplete

images_to_samples.py
remtav: your absolutely right. The approach we had is not robust. moving the old directory seems like a good fix, again, just for debugging purposes.

CharlesAuthier (commented):
Good PR, just change 2-3 things, and can you add in the `conf/README.md` the documentation for `max_used_ram` and `max_used_perc` please (and can you also but what do we need to put in `max_used_ram` to use full ram without knowing the memory that we have)

remtav (approved):

config/config_template.yaml
CharlesAuthier: @remtav

images_to_samples.py
bstdenis: New proposition now moves the old directory...

remtav (approved):

images_to_samples.py
```diff
@@ -326,7 +329,8 @@ def main(params):
 
     final_samples_folder = None
 
-    sample_path_name = f'samples{samples_size}_overlap{overlap}_min-annot{min_annot_perc}_{num_bands}bands'
+    run_name = get_key_def('mlflow_run_name', params['global'], default='gdl')
+    sample_path_name = f'samples{samples_size}_overlap{overlap}_min-annot{min_annot_perc}_{num_bands}bands_{run_name}'
```
remtav, line 333: In our case, an mlflow run is a single training, right? Since multiple runs (ex.: "test13_unet", "test13_deeplabv3", etc.) can share the same samples, should the samples really be named after the run? It seems to me the experiment name might be more relevant, no?

bstdenis: I am also struggling with this, I started using the same run_name when samples are shared. I agree this is am ambiguous use of the concept of mlflow run. I wouldn't mind changing this to the experiment name...

remtav (dismissed):

config/README.md
```diff
@@ -69,12 +73,20 @@ global:
...
+
+- **`mlflow_run_name` :** Run name in *mlflow* and preprocessing identification tag.
+
 - **`bucket_name` (Optional) :** Name of the S3 bucket where the data is stored. Leave blank if using local files.
 
 - **`task` :** Task to perform, either segmentation or classification, but classification is no longer supported.
 
 - **`num_gpus` :** Number of **GPUs** that you want to use, `0` will use the **CPU**.
 
+- **`max_used_ram` :** Maximum used **GPUs** memory (MB) in order to use it.
```
mpelchat04, line 86: Is it the max memory available for the run or the max memory already used by another process that won't be taking into account when searching for available GPUs?  
And how does the `max_used_perc` interact with this parameter?  Do both conditions have to be met?

mpelchat04 (dismissed):

config/README.md
bstdenis: max already used. both conditions are checked and have to be met. I reworded in the last commit, is that better?

mpelchat04: Excellent. Thanks!

mpelchat04 (approved):

CharlesAuthier (approved):

victorlazio109 (approved):

PR #192, opened 2021-03-05
Major changes to inference.py and other minor changes

victorlazio109:


inference.py
```diff
@@ -323,61 +284,77 @@ def main(params: dict):
...
-    params = read_parameters(args.param_file)
+
+    if args.param:
+        params = read_parameters(args.param_file)
+    elif args.input:
+        model = Path(args.input[0])
+        image = args.input[1]
+
+        checkpoint = load_checkpoint(model)
+        params = checkpoint['params']
```
remtav, line 350: maybe raise an error with custom message if key "params" not in model checkpoint. e.g.: "No parameters found in checkpoint. Use GDL version 1.3 or more."

utils/augmentation.py
```diff
@@ -355,13 +356,19 @@ def __call__(self, sample):
...
 
         map_img = None
-        if 'map_img' in sample.keys():  # This can also be used in inference.
-            map_img = np.int64(sample['map_img'])
-            map_img = torch.from_numpy(map_img)
+        if 'map_img' in sample.keys():
+            if sample['map_img'] is not None:  # This can also be used in inference.
+                map_img = np.int64(sample['map_img'])
+                map_img[map_img > self.num_classes] = 0
+                map_img = torch.from_numpy(map_img)
```
remtav, line 373: doesn't this affect the values at 255 that could be passed as "ignore_index" ? I know you're not a fan or ignore_index, @victorlazio109...

utils/metrics.py
```diff
@@ -114,82 +111,38 @@ def __init__(self, label, pred, num_classes):
...
-            c_label = self.label.ravel() == i
-            if c_label.sum() == 0:
-                classes.append(np.nan)
-                continue
-            c_pred = self.pred.ravel()== i
+            c_label = self.label == i
+            c_pred = self.pred == i
+            # if c_label.sum() == 0 and c_pred.sum() == 0:
+            #     classes.append(np.nan)
+            # else:
```
remtav, line 120: what will happen to these commented lines?

inference.py
```diff
@@ -10,169 +10,127 @@
...
-                    params, dataset="tst", type='totensor'
-                )
-                sub_images = padded_img[row:row+sample_size, col:col+sample_size, :]
-                sample['sat_img'] = sub_images
-                sample = totensor_transform(sample)
-                inputs = sample['sat_img'].unsqueeze_(0)
-                inputs = inputs.to(device)
+    WINDOW_SPLINE_2D = _window_2D(window_size=sample_size, power=2.0)
+    WINDOW_SPLINE_2D = torch.as_tensor(np.moveaxis(WINDOW_SPLINE_2D, 2, 0), ).type(torch.float)
+    WINDOW_SPLINE_2D = WINDOW_SPLINE_2D.to(device)
```
remtav, line 47: is smoothing being done systematically now? is there any interest in being able to infer without it?

inference.py
```diff
@@ -10,169 +10,127 @@
...
+                    x_4, y_4 = x_1, y_3
+                    geom = Polygon([(x_1, y_1), (x_2, y_2), (x_3, y_3), (x_4, y_4)])
+                    feature['geometry'].append(geom)
+                    feature['length'].append(geom.length)
+                    feature['pointx'].append(geom.centroid.x)
+                    feature['pointy'].append(geom.centroid.y)
+                    feature['area'].append(geom.area)
+                    cnt += 1
+        gdf = gpd.GeoDataFrame(feature, crs=input_image.crs)
+        gdf.to_crs(crs="EPSG:4326", inplace=True)
```
remtav, line 132: could/should this be extracted as a standalone function?

inference.py
```diff
@@ -309,9 +267,12 @@ def main(params: dict):
         with tqdm(list_img, desc='image list', position=0) as _tqdm:
             for info in _tqdm:
                 img_name = Path(info['tif']).name
-                local_gpkg = info['gpkg']
+                local_gpkg = info['gpkg'] if 'gpkg' in info.keys() else None
```
remtav, line 270: I would right away convert it to Path:
`Path(info['gpkg']) if 'gpkg' in info.keys() else None`

inference.py
```diff
@@ -309,9 +267,12 @@ def main(params: dict):
...
                 if local_gpkg:
                     local_gpkg = Path(local_gpkg)
+                    gpkg_name = Path(local_gpkg).stem
+                else:
+                    gpkg_name = None
```
remtav, line 275: Related to previous comment. I like one-liners for things like this:
`gpkg_name = local_gpkg.stem is local_gpkg else None`

inference.py
```diff
@@ -323,61 +284,77 @@ def main(params: dict):
...
-
                     with rasterio.open(inference_image, 'w+', **inf_meta) as dest:
                         dest.write(pred)
-        log_artifact(working_folder)
+        if len(gdf_) >= 1:
+            assert len(gdf_) == len(gpkg_name_), 'benchmarking unable to complete'
+            all_gdf = pd.concat(gdf_)
+            all_gdf.reset_index(drop=True, inplace=True)
+            gdf_x = gpd.GeoDataFrame(all_gdf)
+            gdf_x.to_file(working_folder.joinpath("benchmark.gpkg"), driver="GPKG", index=False)
```
remtav, line 326: could you add one or two comments on what is going on here. I have a vague idea but since the gdf is populated higher up in the for loop, I think comments would be useful

utils/augmentation.py
victorlazio109: Yes! for the following reasons:
- we allow un-wanted values to pass through at data creation which introduces inconsistencies. At the point of feeding data into the model, I want to be sure my values are correct in the right order. Passing labels as [0, 1, 2, 3] vs [1, 2, 3, 255] (255 is greater than the num of categorical classes we define or are looking to detect)
- ignore_index is only accounted for in the loss function that supports that flag. There are losses that expect only correct number of categorical classes there is no support for ignore_index
- I think there is simplicity in passing only the required num of classes
This might not be the best implementation, however, if we must pass an --ignore_index flag to our loss function we can certainly isolate it and it is best practice to have the class ignored within the range of the classes you are trying to detect.

utils/metrics.py
victorlazio109: I would delete them, I think the current behavior is preferred confirmed by Pierre

inference.py
```diff
@@ -10,169 +10,127 @@
...
-                sub_images = padded_img[row:row+sample_size, col:col+sample_size, :]
-                sample['sat_img'] = sub_images
-                sample = totensor_transform(sample)
-                inputs = sample['sat_img'].unsqueeze_(0)
-                inputs = inputs.to(device)
+    WINDOW_SPLINE_2D = _window_2D(window_size=sample_size, power=2.0)
+    WINDOW_SPLINE_2D = torch.as_tensor(np.moveaxis(WINDOW_SPLINE_2D, 2, 0), ).type(torch.float)
+    WINDOW_SPLINE_2D = WINDOW_SPLINE_2D.to(device)
 
-                if inputs.shape[1] == 4 and any("module.modelNIR" in s for s in model.state_dict().keys()):
```
CharlesAuthier, line 78: did you test for the NIR?

inference.py
victorlazio109: I think it works well now, there are no artifacts introduced and it takes advantage of GPU resources.

inference.py
victorlazio109: opened to syntax corrections!

utils/README.md
```diff
@@ -35,7 +35,7 @@ precision= true positives /
 recall= true positives /
         true positives + false negatives
 ```
-
+<!--
```
CharlesAuthier, line 38: Why?

utils/metrics.py
```diff
@@ -114,82 +111,38 @@ def __init__(self, label, pred, num_classes):
...
     @staticmethod
-    def jaccard(label, pred):
+    def iou(label, pred):
         '''
         :return: IOU
         '''
-        both = np.logical_and(label, pred)
-        either = np.logical_or(label, pred)
-        ji = (np.sum(both) + min_val) / (np.sum(either) + min_val)
+        intersection = (pred & label).sum()
```
CharlesAuthier, line 134: Is it faster to do it like that or with numpy ?

inference.py
victorlazio109: better code syntax, I just need to check if the truth evaluation works with Path.

inference.py
victorlazio109: basically, we are concatenating all the benchmarks for each individual image into one larger table. I would add comments!

inference.py
victorlazio109: okay, more explicit than the inbuilt error message

utils/augmentation.py
remtav: Get it. It's true, not all losses have ignore_index implemented and I see how that could cause problems. Now, we'd need to see what would happen to parts in the code where where ignore_index is used. For example, in images_to_samples.py, we overwrite label values with ignore_index value if pixel corresponds to nodata value in image:
https://github.com/NRCan/geo-deep-learning/blob/develop/images_to_samples.py#L439

If there are large parts of the image that nodata values, then it seems logical to me that we don't want the model to learn on prediction in that area, thus that overwritting... I'd like to talk it over with you and other teammembers. Maybe at our meeting tomorrow am.

prod_requirements.txt
```diff
@@ -0,0 +1,112 @@
...
+sqlparse==0.3.1
+tabulate==0.8.7
+threadpoolctl==2.1.0
+timm==0.2.1
+toolz==0.10.0
+torch==1.4.0
+torchvision==0.5.0
+tornado==6.0.3
+tqdm==4.42.0
+ttach==0.0.3
```
remtav, line 109: is this new dependency added to .travis.yml?

victorlazio109: not yet, I will add in the next commit

utils/README.md
victorlazio109: I noticed lots of weird behaviors with other metrics specifically with multiclass evaluation. Only dice and IoU have been tested extensively, so I decided to exclude them for now.

inference.py
victorlazio109: aaah yes, almost forgot to include it in inference. Thanks!

inference.py
```diff
@@ -323,61 +300,79 @@ def main(params: dict):
...
+
+    parser.add_argument('-p', '--param', metavar='yaml_file', nargs=1,
                         help='Path to parameters stored in yaml')
+    parser.add_argument('-i', '--input', metavar='model_pth img_dir', nargs=2,
+                        help='model_path and image_dir')
     args = parser.parse_args()
-    params = read_parameters(args.param_file)
+
+    if args.param:
+        params = read_parameters(args.param_file)
```
remtav, line 360: After testing, I think line 360 should be 
`params = read_parameters(args.param[0])`
not 
`params = read_parameters(args.param_file)`

inference.py
```diff
@@ -323,61 +300,79 @@ def main(params: dict):
...
+        # log_artifact(working_folder)
     time_elapsed = time.time() - since
     print('Inference and Benchmarking completed in {:.0f}m {:.0f}s'.format(time_elapsed // 60, time_elapsed % 60))
 
+
 if __name__ == '__main__':
     print('\n\nStart:\n\n')
-    parser = argparse.ArgumentParser(description='Inference and Benchmark on images using trained model')
-    parser.add_argument('param_file', metavar='file',
+    parser = argparse.ArgumentParser(usage="%(prog)s [-h] [YAML] [-i MODEL IMAGE] ",
```
remtav, line 350: can you update this text? the -p parameter is not showing. Instead, there's a -h parameter that isn't used in this version.

inference.py
victorlazio109: yeesh! an oversight. corrected!

remtav (dismissed):

remtav (approved):

utils/geoutils.py
```diff
@@ -180,4 +181,4 @@ def get_key_recursive(key, config):
     if isinstance(val, (dict, collections.OrderedDict)):
         assert len(key) > 1, "missing keys to index metadata subdictionaries"
         return get_key_recursive(key[1:], val)
-    return val
+    return int(val)
```
remtav, line 184: Isn't this value already casted to int here: https://github.com/NRCan/geo-deep-learning/blob/develop/utils/geoutils.py#L31? 
If casting to int is necessary here, I would suggest doing this small modification [in the list-ids function at line 25](at https://github.com/NRCan/geo-deep-learning/blob/develop/utils/geoutils.py#L25), rather that in the generic get_key_recursive. If, in the future, this utility funciton is used to return anything else but an int, `int(val)` could raise an undesired exception. I'd keep `return val` here. Up to you if you want to integrate it or not.

remtav (approved):

utils/geoutils.py
victorlazio109: I would skip this change for now to merge. I would note this and fix in a subsequent PR

environment.yml
```diff
@@ -0,0 +1,221 @@
+name: geo_deep_env
```
CharlesAuthier, line 1: love the idea

CharlesAuthier (approved):
Sorry to not telling you that sooner, if you remouve the `requirement.txt` can you change the main `README.md` in the section Requirement. and meaby add the way to use the environement yaml in the installation section, but we can do that in the next PR too I approve.

environment.yml
victorlazio109: yes! am aware Yves C made those changes. I would merge.

PR #197, opened 2021-03-25
Implement python's logging library, code maintenance and refactoring, bug fixes 

remtav:
expose major yaml parameters at top of main() at top of our main scripts (images_to_samples, train_..., inference)
images_to_samples.py: fix target_background_percent when all vals are ignore_index (bugfix)
use python's logging package
implement single-class samples creation from multi-class vector data (gpkg) (1st version to be improved)
minor bug fixes


images_to_samples.py
```diff
@@ -124,28 +129,28 @@ def add_to_datasets(dataset,
...
     # adds pixel count to pixel_classes dict for each class in the image
-    for key, value in enumerate(np.bincount(target.clip(min=0).flatten())):
-        cls_keys = dict_classes.keys()
-        if key in cls_keys:
-            dict_classes[key] += value
-        elif key not in cls_keys and value > 0:
-            raise ValueError(f"A class value was written ({key}) that was not defined in the classes ({cls_keys}).")
-
-    return val
+    class_vals, counts = np.unique(target, return_counts=True)
```
victorlazio109, line 146: neat!

images_to_samples.py
```diff
@@ -124,28 +129,28 @@ def add_to_datasets(dataset,
...
+    for i in range(len(class_vals)):
+        class_val = class_vals[i]
+        count = counts[i]
+        dict_classes[class_val] += count
+        if class_val not in dict_classes.keys():
+            raise ValueError(f'Sample contains value "{class_val}" not defined in the classes ({dict_classes.keys()}).')
+    return to_val_set
```
victorlazio109, line 154: neat!

images_to_samples.py
```diff
@@ -225,8 +230,9 @@ def samples_preparation(in_img_array,
                     padding = pad_diff(target_row, target_col, sample_size,
                                        sample_size)  # array, actual height, actual width, desired size
                     target = pad(target, padding, fill=dontcare)
-                u, count = np.unique(target, return_counts=True)
-                target_background_percent = round(count[0] / np.sum(count) * 100 if 0 in u else 0, 1)
+                backgr_ct = np.sum(target == 0)
+                backgr_ct += np.sum(target == dontcare)
+                target_background_percent = round(backgr_ct / target.size * 100, 1)
```
victorlazio109, line 236: Suggestion/thoughts: different from target_background percent/min_annot_percent I think it would be 'a nice to have' if we are able to keep 20/30%(even less %5) of samples containing all background while disposing of 80/70%

images_to_samples.py
```diff
@@ -309,72 +316,109 @@ def main(params):
...
+    metadata = None
+    targ_ids = get_key_def('target_ids', params['sample'], None, expected_type=List)
+    class_prop = get_key_def('class_proportion', params['sample']['sampling_method'], None, expected_type=dict)
+    mask_reference = get_key_def('mask_reference', params['sample'], default=False, expected_type=bool)
 
-    experiment_name = get_key_def('mlflow_experiment_name', params['global'], default='gdl-training')
-    sample_path_name = (f'samples{samples_size}_overlap{overlap}_min-annot{min_annot_perc}_'
-                        f'{num_bands}bands_{experiment_name}')
+    # add git hash from current commit to parameters if available. Parameters will be saved to hdf5s
+    params['global']['git_hash'] = get_git_hash()
```
victorlazio109, line 359: Well Done!

images_to_samples.py
```diff
@@ -309,72 +316,109 @@ def main(params):
...
-            shutil.move(samples_folder, f'{str(samples_folder)}_{uuid.uuid4().hex[0:6]}')
+            last_mod_time_suffix = datetime.fromtimestamp(smpls_dir.stat().st_mtime).strftime('%Y%m%d-%H%M%S')
+            shutil.move(smpls_dir, data_path.joinpath(f'{str(smpls_dir)}_{last_mod_time_suffix}'))
         else:
-            raise FileExistsError(f'Data path exists: {samples_folder}. Remove it or use a different experiment_name.')
-    else:
-        tqdm.write(f'Writing samples to {samples_folder}')
-    Path.mkdir(samples_folder, exist_ok=False)  # TODO: what if we want to append samples to existing hdf5?
-    tqdm.write(f'Samples will be written to {samples_folder}\n\n')
+            raise FileExistsError(f'Data path exists: {smpls_dir}. Remove it or use a different experiment_name.')
```
victorlazio109, line 379: The folder may exist and yet be empty, maybe raise a warning instead?

images_to_samples.py
```diff
@@ -309,72 +316,109 @@ def main(params):
...
+    Path.mkdir(smpls_dir, exist_ok=False)  # TODO: what if we want to append samples to existing hdf5?
 
-    tqdm.write(f'\nSuccessfully read csv file: {Path(csv_file).stem}\n'
-               f'Number of rows: {len(list_data_prep)}\n'
-               f'Copying first entry:\n{list_data_prep[0]}\n')
-    ignore_index = get_key_def('ignore_index', params['training'], -1)
-    meta_map, metadata = get_key_def("meta_map", params["global"], {}), None
+    import logging.config  # See: https://docs.python.org/2.4/lib/logging-config-fileformat.html
```
victorlazio109, line 382: hmmn, ansy when i see python 2, we are at python 3.x for GDL. it could be the functionality is the same in python3?

inference.py
```diff
@@ -225,47 +227,83 @@ def main(params: dict):
...
+            Path(mlflow_uri).mkdir()
+        set_tracking_uri(mlflow_uri)
+        exp_name = get_key_def('mlflow_experiment_name', params['global'], default='gdl-inference', expected_type=str)
+        set_experiment(f'{exp_name}/{working_folder.name}')
+        run_name = get_key_def('mlflow_run_name', params['global'], default='gdl', expected_type=str)
+        start_run(run_name=run_name)
+        log_params(params['global'])
+        log_params(params['inference'])
+
+    # AWS
```
victorlazio109, line 276: all things logging and mlflow in inference should be optional, maybe we can implement a logging flag here or use the existing debug flag (default=false). The usage of inference.py is not limited to only developers.

images_to_samples.py
bstdenis: Why would there be an empty folder sitting there? Doesn't that point to something gone wrong previously that needs to be cleaned up?

train_segmentation.py
```diff
@@ -197,23 +243,28 @@ def train(train_loader,
...
-    vis_at_train = get_key_def('vis_at_train', vis_params['visualization'], False)
-    vis_batch_range = get_key_def('vis_batch_range', vis_params['visualization'], None)
 
     with tqdm(train_loader, desc=f'Iterating train batches with {device.type}') as _tqdm:
         for batch_index, data in enumerate(_tqdm):
             progress_log.open('a', buffering=1).write(tsv_line(ep_idx, 'trn', batch_index, len(train_loader), time.time()))
 
-            inputs = data['sat_img'].to(device)
-            labels = data['map_img'].to(device)
+            try:  # For HPC when device 0 not available. Error: RuntimeError: CUDA error: invalid device ordinal
```
victorlazio109, line 259: why? the device type is already inferred and it's being passed into this function. should we be changing that in this try and except statement?

utils/geoutils.py
```diff
@@ -125,7 +128,10 @@ def vector_to_raster(vector_file, input_image, out_shape, attribute_name, fill=0
     lst_vector_tuple = lst_ids(list_vector=lst_vector, attr_name=attribute_name, target_ids=target_ids,
                                merge_all=merge_all)
 
-    if merge_all:
+    if not lst_vector_tuple:  # if no vectors found, return None and prevent error in rasterize function below.
```
victorlazio109, line 131: yaaay! single class

victorlazio109 (commented):
very elaborate, I like! just a few comments to address

train_segmentation.py
```diff
@@ -78,40 +93,61 @@ def create_dataloader(samples_folder, batch_size, num_devices, params):
...
+        # get max ram for smallest gpu
+        smallest_gpu_ram = min(gpu_info['max_ram'] for _, gpu_info in devices_dict.items())
+        # rule of thumb to determine eval batch size based on approximate max pixels a gpu can handle during evaluation
+        max_pix_per_mb_gpu = 150  # TODO: this value may need to be finetuned
+        pix_per_mb_gpu = (batch_size / len(devices_dict.keys()) * sample_size ** 2) / smallest_gpu_ram
+        if pix_per_mb_gpu >= max_pix_per_mb_gpu:
+            eval_batch_size = int(round(smallest_gpu_ram * max_pix_per_mb_gpu / sample_size**2, len(devices_dict.keys())))
+            eval_batch_size = 1 if eval_batch_size < 1 else eval_batch_size
+            logging.warning(f'Validation and test batch size downgraded from {batch_size} to {eval_batch_size} '
+                            f'based on max ram of smallest GPU available')
```
remtav, line 129: this will be presented and discussed.

train_segmentation.py
```diff
@@ -124,13 +160,15 @@ def get_num_samples(samples_path, params):
...
-                        samples_weight = compute_sample_weight('balanced', weights)
+
+        with h5py.File(samples_path.joinpath(f"{i}_samples.hdf5"), "r") as hdf5_file:
+            if i == 'trn':
+                for x in range(num_samples[i]):
+                    label = hdf5_file['map_img'][x]
+                    label = np.where(label == dontcare, 0, label)
+                    unique_labels = np.unique(label)
+                    weights.append(''.join([str(int(i)) for i in unique_labels]))
+                    samples_weight = compute_sample_weight('balanced', weights)
```
remtav, line 171: small bugfix when using num_samples

train_segmentation.py
```diff
@@ -545,20 +703,17 @@ def main(params, config_path):
                         'model': state_dict,
                         'best_loss': best_loss,
                         'optimizer': optimizer.state_dict()}, filename)
-            if epoch == 0:
-                log_artifact(filename)
```
remtav, line 549: why were we saving the checkpoing at epoch=0 in mlflow? @mpelchat04 @victorlazio109 ?

utils/logging.conf
```diff
@@ -0,0 +1,33 @@
...
+
+[handler_consoleHandler]
+class=StreamHandler
+level=INFO
+formatter=fmt
+args=(sys.stdout,)
+
+[handler_fileHandler]
+class=FileHandler
+level=INFO
```
remtav, line 22: this can be modified with time according to specific needs

utils/utils.py
```diff
@@ -82,28 +85,33 @@ def list_s3_subfolders(bucket, data_path):
...
+                used_ram = mem.used / (1024 ** 2)
+                max_ram = mem.total / (1024 ** 2)
                 if debug:
-                    print(f'GPU RAM used: {round(mem.used/(1024**2), 1)} | GPU % used: {res.gpu}')
-                if round(mem.used/(1024**2), 1) <  max_used_ram and res.gpu < max_used_perc:
-                    lst_free_devices.append(i)
-                if len(lst_free_devices) == number_requested:
+                    logging.info(f'GPU RAM used: {used_ram:.0f}/{max_ram:.0f} MiB\nGPU % used: {res.gpu}')
+                if used_ram/max_ram < max_used_ram_perc and res.gpu < max_used_perc:
+                    lst_free_devices[i] = {'used_ram_at_init': used_ram, 'max_ram': max_ram}
```
remtav, line 109: returning info about gpu's ram back to main()

train_segmentation.py
remtav: Often, on HPC, our get_device_ids returns that gpu 2 and 3 are available but I get a runtime error when using those values instead of 0, 1. Maybe the fact that the machine it's being executed is a virtual machine. Not sure exactly why, but I got into this problem again recently since the ".to(device)" is used in multiple place in our code now.

images_to_samples.py
remtav: I guess... I tested and it works well...

images_to_samples.py
remtav: yes. This was added by @bstdenis in PR #181

inference.py
```diff
@@ -225,47 +228,86 @@ def main(params: dict):
...
+    # SETTING OUTPUT DIRECTORY
     working_folder = Path(params['inference']['state_dict_path']).parent.joinpath(f'inference_{num_bands}bands')
-    num_devices = params['global']['num_gpus'] if params['global']['num_gpus'] else 0
     Path.mkdir(working_folder, parents=True, exist_ok=True)
-    print(f'Inferences will be saved to: {working_folder}\n\n')
 
+
+    # mlflow logging
+    mlflow_uri = get_key_def('mlflow_uri', params['global'], default=None, expected_type=str)
+    if mlflow_uri:
```
remtav, line 258: this is a quick fix to integrate your comment, @victorlazio109 
is it ok to use the existence of mlflow_uri as a flag to activate all logging?

train_segmentation.py
```diff
@@ -403,65 +481,167 @@ def main(params, config_path):
...
+    logging.info(f'GPUs devices available: {gpu_devices_dict}')
+    num_devices = len(gpu_devices_dict.keys())
+    device = torch.device(f'cuda:{list(gpu_devices_dict.keys())[0]}' if gpu_devices_dict else 'cpu')
+
+    logging.info(f'Creating dataloaders from data in {samples_folder}...\n')
+
+    # overwrite dontcare values in label if loss is not lovasz or crossentropy. FIXME: hacky fix.
+    dontcare2backgr = True if loss_fn not in ['Lovasz', 'CrossEntropy', 'OhemCrossEntropy'] else False
+    logging.warning(f'Dontcare is not implemented for loss function "{loss_fn}". '
+                    f'Dontcare values ({dontcare_val}) in label will be replaced with background value (0)')
```
remtav, line 604: @victorlazio109 what do you think of this as a hacky fix for loss functions that don't implement ignore_index?
see where this parameter ends up in augmentations.py --> ToTensorTarget

images_to_samples.py
remtav: definitely. Could you open a "feature enhancement" issue?

images_to_samples.py
remtav: @victorlazio109 I agree with @bstdenis to keep the FileExistsError. It will be raised at the very beginning of images_to_samples so not too frustrating. Also, for your information, behavior during training is different. During training, the old directory is renamed and no errors are raised.

train_segmentation.py
```diff
@@ -403,65 +483,167 @@ def main(params, config_path):
...
+
+    # automatic model naming with unique id for each training
+    model_id = config_path.stem
+    output_path = samples_folder.joinpath('model') / model_id
+    if output_path.is_dir():
+        last_mod_time_suffix = datetime.fromtimestamp(output_path.stat().st_mtime).strftime('%Y%m%d-%H%M%S')
+        archive_output_path = samples_folder.joinpath('model') / f"{model_id}_{last_mod_time_suffix}"
+        shutil.move(output_path, archive_output_path)
+    output_path.mkdir(parents=True, exist_ok=False)
+    shutil.copy(str(config_path), str(output_path))  # copy yaml to output path where model will be saved
```
remtav, line 575: @victorlazio109 see how this is different than images_to_samples.py. No exceptions raised.

images_to_samples.py
victorlazio109: Agreed.

train_segmentation.py
victorlazio109: seen, thanks

train_segmentation.py
victorlazio109: I guess it's fine. I would let others comment on this.

bstdenis (dismissed):

bstdenis (dismissed):

models/model_choice.py
```diff
@@ -51,14 +56,22 @@
...
         }},
 }
+try:
+    lm_smp['manet_pretrained'] = {
+        # https://ieeexplore.ieee.org/abstract/document/9201310
+        'fct': smp.MAnet, 'params': {
+            'encoder_name': 'resnext50_32x4d'}}
+except AttributeError:
+    logging.exception("Couldn't load MAnet from segmentation models pytorch package. Check installed version")
+
```
remtav, line 66: MAnet not implemented in older versions

models/model_choice.py
```diff
@@ -99,81 +125,76 @@ def set_hyperparameters(params, num_classes, model, checkpoint, dontcare_val):
...
+        pretrained: bool = True,
+        dropout_prob: float = False,
+        loss_fn: str = None,
+        optimizer: str = None,
+        class_weights: Sequence = None,
+        net_params=None,
+        conc_point: str = None,
+        coordconv_params=None,
+        inference_state_dict: str = None):
     """Define the neural net"""
```
remtav, line 162: A lot more arguments are passed in net() function rather than only "params" dictionary.

models/model_choice.py
```diff
@@ -191,14 +212,18 @@ def net(net_params, num_channels, inference=False):
...
-                #conv1 = model.backbone._modules['conv1'].weight.detach().numpy()
-                #depth = np.random.uniform(low=-1, high=1, size=(64, 1, 7, 7))
-                #conv1 = np.append(conv1, depth, axis=1)
-                #conv1 = torch.from_numpy(conv1).float()
-                #model.backbone._modules['conv1'].weight = nn.Parameter(conv1, requires_grad=True)
+                # conv1 = model.backbone._modules['conv1'].weight.detach().numpy()
+                # depth = np.random.uniform(low=-1, high=1, size=(64, 1, 7, 7))
+                # conv1 = np.append(conv1, depth, axis=1)
+                # conv1 = torch.from_numpy(conv1).float()
+                # model.backbone._modules['conv1'].weight = nn.Parameter(conv1, requires_grad=True)
```
remtav, line 219: Keeping this since our test13 model with deeplabv3 seems to have used this. Both solutions work but are not interchangeable. A model trained with this approach cannot be built afterwars with LayersEnsemble approach (e.g. for inference)

train_segmentation.py
```diff
@@ -78,40 +102,81 @@ def create_dataloader(samples_folder, batch_size, num_devices, params):
...
+    num_workers = len(gpu_devices_dict.keys()) * 4 if len(gpu_devices_dict.keys()) > 1 else 4
 
     samples_weight = torch.from_numpy(samples_weight)
     sampler = torch.utils.data.sampler.WeightedRandomSampler(samples_weight.type('torch.DoubleTensor'),
                                                              len(samples_weight))
+    eval_batch_size = batch_size
+
+    if gpu_devices_dict and calc_eval_bs:
+        max_pix_per_mb_gpu = 280  # TODO: this value may need to be finetuned
+        calc_eval_batchsize(gpu_devices_dict, batch_size, sample_size, max_pix_per_mb_gpu)
```
remtav, line 134: this should be eval_batch_size = calc_eval_batchsize(...)

train_segmentation.py
```diff
@@ -78,40 +102,81 @@ def create_dataloader(samples_folder, batch_size, num_devices, params):
...
                                 drop_last=True)
-    tst_dataloader = DataLoader(tst_dataset, batch_size=1, num_workers=num_workers, shuffle=False,
+    tst_dataloader = DataLoader(tst_dataset, batch_size=eval_batch_size, num_workers=num_workers, shuffle=False,
                                 drop_last=True) if num_samples['tst'] > 0 else None
 
     return trn_dataloader, val_dataloader, tst_dataloader
 
 
-def get_num_samples(samples_path, params):
+def calc_eval_batchsize(gpu_devices_dict: dict, batch_size: int, sample_size: int, max_pix_per_mb_gpu: int = 280):
```
remtav, line 146: Again, this may not reach consensus. Trying to automatically calculate evaluation batch size based on an approximate "maximum pixels per Mb of GPU" (this also depends on models. I overestimated the value to make sure it passes even with large models such as deeplabv3)

losses/boundary_loss.py
```diff
@@ -16,6 +31,10 @@ def __init__(self, theta0=19, theta=19, ignore_index=None, weight=None):
         self.theta = theta
         self.ignore_index = ignore_index
         self.weight = weight
+        if self.ignore_index:
+            logging.error(f'Ignore_index not implemented for Boundary Loss. Got ignore_index "{ignore_index}"')
+        if self.ignore_index:
+            logging.error(f'Class weights not implemented for Boundary Loss. Got class weights "{weight}"')
```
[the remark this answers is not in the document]
remtav: Corrected. One of the "if" statements should've been "if self.weight"

CharlesAuthier (approved):

train_segmentation.py
mpelchat04: good question. I don't think it's relevant to do so.

mpelchat04 (dismissed):

CharlesAuthier (approved):

mpelchat04 (approved):

CharlesAuthier (approved):

victorlazio109 (approved):

PR #201, opened 2021-05-31
[WIP] Add hyperopt template

bstdenis:
Using hyperopt with GDL for hyperparameter optimization

ToDo:
- [ ] Make it application agnostic. Currently has some code specific to flood mapping

### DIFF
--- a/gdl_hyperopt_template.py
+++ b/gdl_hyperopt_template.py
@@ -1,0 +1,152 @@
+"""Hyperparamater optimization for GDL using hyperopt
+
+This is a template for using hyperopt with GDL. The my_space variable currently needs to
+be modified here, as well as GDL config modification logic within the objective_with_args
+function.
+
+"""
+
+import argparse
+from pathlib import Path
+import pickle
+from functools import partial
+import pprint
+
+import mlflow
+import torch
+# ToDo: Add hyperopt to GDL requirements
+from hyperopt import fmin, tpe, hp, Trials, STATUS_OK
+
+from utils.readers import read_parameters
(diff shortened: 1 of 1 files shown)


gdl_hyperopt_template.py
```diff
@@ -0,0 +1,152 @@
...
+
+This is a template for using hyperopt with GDL. The my_space variable currently needs to
+be modified here, as well as GDL config modification logic within the objective_with_args
+function.
+
+"""
+
+import argparse
+from pathlib import Path
+import pickle
```
ymoisan, line 11: See https://docs.python.org/3/library/pickle.html.  And https://medium.com/swlh/object-serialization-and-deserialization-in-python-5fad3c2970a4 (Marshmallow).  Maybe even more "strategic" would be to start using Google's [protobuf](https://pypi.org/project/protobuf-serialization/).  This is how Mapbox serializes their vector tiles IIRC

victorlazio109 (approved):

CharlesAuthier (approved):

gdl_hyperopt_template.py
```diff
@@ -0,0 +1,152 @@
...
+    pprint.pprint(trials.results)
+    for key, val in best.items():
+        if my_space[key].name == 'switch':
+            best[key] = my_space[key].pos_args[val+1].obj
+    pprint.pprint(best)
+    print(trials.best_trial['result'])
+    trials_to_csv(trials)
+
+
+if __name__ == '__main__':
```
remtav, line 146: Is there a usage example with hyperopt somewhere in the doc?

PR #202, opened 2021-07-05
Add non background IoU

bstdenis:


mpelchat04 (approved):

CharlesAuthier (approved):

utils/metrics.py
```diff
@@ -89,6 +92,18 @@ def iou(pred, label, batch_size, num_classes, metric_dict, only_present=True):
         iou = (intersection + min_val) / (union + min_val)  # minimum value added to avoid Zero division
         ious.append(iou)
         metric_dict['iou_' + str(i)].update(iou.item(), batch_size)
+
+    # Add overall non-background iou metric
+    c_label = label == 1
+    c_pred = pred == 1
+    for i in range(2, num_classes):
```
remtav, line 99: shouldn't this be :
`for i in range (1, num_classes)` ?
We only want to exclude 0 values, right?

bstdenis: Class 1 is initialized above before the loop, and inside the loop it does logical_or with those. So this is all classes except 0.

remtav: Ah! I understand. Since we're not interested in the perclass iou (already computed prior to this iou), wouldn't it be simpler to remove the for loop and use something like:
`c_label = label >= 1
c_pred = pred >= 1`
In this case, you could make sure there aren't more values in prediction and label than the num_classes. Your approach has the advantage of computing only on class values, but wouldn't raise a warning if values above num_classes exist. This assessment may be done elsewhere in the code though.

bstdenis: Ah yes, that would work the same and is simpler, want me to change it?

remtav: If you'd like. It may be faster if we are computing iou's hundreds and hundreds of times. Here's another implementation closer to yours (safer if there are any values superior to num_classes-1) :
`c_label = (1 <= label) & (label < num_classes)`
`c_pred = (1 <= pred) & (pred < num_classes)`

Sidenote, I've been looking through numpy docs for a while to find this type of boolean masking with multiple conditions :)

remtav (approved):

mpelchat04 (approved):

PR #206, opened 2021-08-12
Major and Minor updates to GDL

victorlazio109:
Main Highlights
- New sample creation script
-  Major updates to inference.py
- minor fixes across GDL


gdl_hyperopt_HPC.py
```diff
@@ -0,0 +1,163 @@
+"""Hyperparamater optimization for GDL using hyperopt
+
+This is a template for using hyperopt with GDL. The my_space variable currently needs to
+be modified here, as well as GDL config modification logic within the objective_with_args
+function.
+
+"""
```
CharlesAuthier, line 7: Can you put a section in the `README.md` on how to launch it pls :)

gdl_hyperopt_template.py
```diff
@@ -21,12 +22,10 @@
 from train_segmentation import main as train_main
 
 # This is the hyperparameter space to explore
-my_space = {'target_size': hp.choice('target_size', [128, 256]),
-            'model_name': hp.choice('model_name', ['unet', 'deeplabv3+_pretrained']),
-            'permanent_water_weight': hp.uniform('permanent_water_weight', 1.0, 10.0),
-            'rivers_weight': hp.uniform('rivers_weight', 1.0, 10.0),
-            'flood_weight': hp.uniform('flood_weight', 1.0, 10.0),
-            'noise': hp.choice('noise', [0.0, 1.0])}
+my_space = {'model_name': hp.choice('model_name', ['unet_pretrained', 'deeplabv3_resnet101']),
```
CharlesAuthier, line 25: Same comment that the HPC one

utils/utils.py
```diff
@@ -71,8 +73,8 @@ def load_from_checkpoint(checkpoint, model, optimizer=None, inference:str=''):
     strict_loading = False if not inference else True
     model.load_state_dict(checkpoint['model'], strict=strict_loading)
     logging.info(f"=> loaded model\n")
-    if optimizer and 'optimizer' in checkpoint.keys():    # 2nd condition if loading a model without optimizer
-        optimizer.load_state_dict(checkpoint['optimizer'], strict=False)
+    # if optimizer and 'optimizer' in checkpoint.keys():    # 2nd condition if loading a model without optimizer
```
CharlesAuthier, line 76: Can we delete it ?

README.md
```diff
@@ -32,7 +32,7 @@ The final step in the process is to assign every pixel in the original image a v
 ## **Requirement**
 This project comprises a set of commands to be run at a shell command prompt.  Examples used here are for a bash shell in an Ubuntu GNU/Linux environment.
 
-- [Python 3.6](https://www.python.org/downloads/release/python-360/), see the full list of dependencies in [requirements.txt](requirements.txt)
+- [Python 3.7.6](https://www.python.org/downloads/release/python-376/), see the full list of dependencies in [environment.yml](environment.yml)
 - [mlflow](https://mlflow.org/)
 - [minicanda](https://docs.conda.io/en/latest/miniconda.html) (highly recommended)
```
ymoisan, line 37: typo : miniconda

gdl_hyperopt_template.py
```diff
@@ -62,20 +61,23 @@ def objective_with_args(hparams, params, config_path):
     """
 
     # ToDo: This is dependent on the specific structure of the GDL config file
-    params['training']['target_size'] = hparams['target_size']
     params['global']['model_name'] = hparams['model_name']
-    # ToDo: Should adjust batch size as a function of model and target size...
-    params['training']['class_weights'] = [1.0, hparams['permanent_water_weight'], hparams['rivers_weight'],
```
ymoisan, line 68: A (hyperopt) template should not have parameters clearly linked to a specific application, e.g. floods like in this case.  "specific structure of the GDL config file" indeed ...  We may need a structure of YAML files as a function of application that a template could make use of.

utils/utils.py
mwesleyj: I think this should stay (although feel free to disagree)

- as if we keep it:
     if `optimizer == None` then it will not change anything
     if `optimizer != None` it will load a better/more accurate/more pre-trained checkpoint 
                                                               ^ for lack of better words
- if we remove it
     we should remove the optimizer var from the params, docs, & return line
                                                                               (the PR's Lines 55, 60, 78)

@victorlazio109 I was having issues with this line too & the way I was able to solve it was to remove the `strict=False` from line 77

my code:
```
    if optimizer and 'optimizer' in checkpoint.keys():    # 2nd condition if loading a model without optimizer
        optimizer.load_state_dict(checkpoint['optimizer'])#, strict=False)
```

- lastly, if it is removed to be able to load checkpoints with diff optimizers..
maybe we could add an `if` statement?
here is a helpful link if this is the case:
https://stackoverflow.com/questions/64301566/how-to-check-if-an-object-is-a-certain-pytorch-optimizer

utils/geoutils.py
```diff
@@ -63,15 +64,18 @@ def clip_raster_with_gpkg(raster, gpkg, debug=False):
     """
     from shapely.geometry import box  # geopandas and shapely become a project dependency only during sample creation
     import geopandas as gpd
+    import fiona
     # Get extent of gpkg data with fiona
     with fiona.open(gpkg, 'r') as src:
+        gpkg_crs = src.crs
+        assert gpkg_crs == raster.crs
```
mwesleyj, line 71: maybe use `utils.verification.assert_crs_match(...)` instead?

as a few times I have noticed they dont always match 
or
I need to check src.crs['init'] instead

mwesleyj: ... I realize src.crs['init'] opens up a whole new can of worms
so mb ignore for now ( I have put on my `# ToDo:` list )

utils/visualization.py
```diff
@@ -123,7 +124,8 @@ def vis(vis_params,
     assert vis_path.parent.is_dir()
     vis_path.mkdir(exist_ok=True)
 
-    if not vis_params['inference_input_path']:  # FIXME: function parameters should not come in as different types if inference or not.
+    if not vis_params[
```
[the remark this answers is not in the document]
victorlazio109: I pulled and merged this change, will fix

gdl_hyperopt_HPC.py
```diff
@@ -0,0 +1,163 @@
...
+from train_segmentation import main as train_main
+
+# This is the hyperparameter space to explore
+# my_space = {'target_size': hp.choice('target_size', [512, 636]),
+#             'model_name': hp.choice('model_name', ['unet_pretrained', 'deeplabv3_resnet101']),
+#             'loss_fn': hp.choice('loss_fn', ['CrossEntropy', 'Lovasz', 'Duo']),
+#             'optimizer': hp.choice('optimizer', ['adam', 'adabound']),
+#             'learning_rate': hp.loguniform('learning_rate', np.log(1e-7), np.log(0.1))}
+
+my_space = {'loss_fn': hp.choice('loss_fn', ['CrossEntropy', 'Lovasz', 'Duo']),
```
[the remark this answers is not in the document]
victorlazio109: this template will be removed, it is experiment specific supposed to live on a private branch

gdl_hyperopt_HPC.py
victorlazio109: will be removed

utils/utils.py
victorlazio109: oh nvm, you commented out strict=False, I will try

utils/geoutils.py
victorlazio109: I will remove that assertion, it isn't meant to be there

utils/geoutils.py
```diff
@@ -80,21 +84,24 @@ def clip_raster_with_gpkg(raster, gpkg, debug=False):
...
+            out_meta.update({"driver": "GTiff",
+                             "height": out_img.shape[1],
+                             "width": out_img.shape[2],
+                             "transform": out_transform})
+            with rasterio.open(out_tif, "w", **out_meta) as dest:
+                print(f"writing clipped raster to {out_tif}")
+                dest.write(out_img)
+            return out_tif
+        except ValueError as e:  # if gpkg's extent outside raster: "ValueError: Input shapes do not overlap raster."
+            # TODO: warning or exception? if warning, except must be set in images_to_samples
```
mwesleyj, line 103: does `utils.readers.image_reader_as_array( ... )` still need to call `clip_raster_with_gpkg( ... )` ?
because it still expects 2 returns (not 1)

mwesleyj: **Warning** (I think)
But this (Lines 102-104) should prob also return something

thinking off the top of my head to fix: I would say 
`return False`
pass thru `process_raster_img` 
after `rst_pth, r_ = process_raster_img(img_pan, gpkg)` test if `rst_path==False` then trigger a flag?
then have the flag `break` out of loop

gdl_hyperopt_template.py
victorlazio109: this would be tackled in a future PR

inference.py
```diff
@@ -43,112 +46,214 @@
...
+        row = img[1]
+        col = img[2]
+        sub_image = img[0]
+        image_metadata = add_metadata_from_raster_to_sample(sat_img_arr=sub_image,
+                                                            raster_handle=input_image,
+                                                            meta_map={},
+                                                            raster_info={})
+
+        sample['metadata'] = image_metadata
+        totensor_transform = augmentation.compose_transforms(params,
```
CharlesAuthier, line 193: `params` is called but never asign

victorlazio109: I investigated this behavior; `params` is not passed to the local function but called globally. Difficult to detect because no error is thrown. This is not ideal and has been fixed with the latest commit. Thanks for pointing it out.

sample_creation.py
```diff
@@ -0,0 +1,569 @@
+import argparse
```
mpelchat04, line 1: Is this script a replacement for Images_to_samples.py? If so, I don't think it's very usefull to have both.

gdl_hyperopt_template.py
```diff
@@ -62,20 +61,23 @@ def objective_with_args(hparams, params, config_path):
...
+    params['training']['optimizer'] = hparams['optimizer']
+    params['training']['learning_rate'] = hparams['learning_rate']
 
     try:
         mlrun = get_latest_mlrun(params)
         run_name_split = mlrun.data.tags['mlflow.runName'].split('_')
-        params['global']['mlflow_run_name'] = run_name_split[0] + f'_{int(run_name_split[1])+1}'
+        params['global']['mlflow_run_name'] = run_name_split[0] + f'_{int(run_name_split[1]) + 1}'
     except:
         pass
```
remtav, line 75: why try/except statement here? If necessary, maybe it should be narrowed down to catch only known errors. Not a priority.

gdl_hyperopt_template.py
```diff
@@ -62,20 +61,23 @@ def objective_with_args(hparams, params, config_path):
...
 
+    if params['global']['model_name'] == "unet_pretrained":
+        params['training']['state_dict_path'] = params['training']['dict_unet']
+    elif params['global']['model_name'] == "deeplabv3_resnet101":
+        params['training']['state_dict_path'] = params['training']['dict_deeplab']
```
remtav, line 80: the dict_[model] parameter comes from where? Is it an output of a previous run? Maybe a comment or two would help here.

images_to_samples.py
```diff
@@ -247,6 +248,7 @@ def samples_preparation(in_img_array,
                 # Stratification bias
                 if (stratd is not None) and (dataset == 'trn'):
                     tile_size = target.size
+                    u, count = np.unique(target, return_counts=True)
```
remtav, line 251: thanks for fixing this bug.

inference.py
```diff
@@ -43,112 +45,223 @@
 logging.getLogger(__name__)
 
 
-def calc_inference_chunk_size(gpu_devices_dict: dict, max_pix_per_mb_gpu: int = 350):
```
remtav, line 46: why remove this function? Did it cause problems?

inference.py
```diff
@@ -43,112 +45,223 @@
...
+    print("")
+    print("Number of features written: {}".format(i))
+
+
+def gen_img_samples(src, chunk_size, *band_order):
+    subdiv = 2.0
+    step = int(chunk_size / subdiv)
+    for row in range(0, src.height, step):
+        for column in range(0, src.width, step):
+            window = Window.from_slices(slice(row, row + chunk_size),
```
remtav, line 125: yeah Window!

inference.py
```diff
@@ -43,112 +45,223 @@
...
+                        layer=layer_name,
+                        schema=feat_schema,
+                        driver='GPKG') as dest:
+            for polygon in polygons:
+                dest.write(polygon)
+    print("")
+    print("Number of features written: {}".format(i))
+
+
+def gen_img_samples(src, chunk_size, *band_order):
```
remtav, line 120: nice generator function. please add docstring to this new function :)

inference.py
```diff
@@ -43,112 +45,223 @@
...
+            if window_array.shape[0] < chunk_size or window_array.shape[1] < chunk_size:
+                window_array = _pad_diff(window_array, window_array.shape[0], window_array.shape[1], chunk_size)
+            window_array = _pad(window_array, chunk_size)
+
+            yield window_array, row, column
 
 
 @torch.no_grad()
-def segmentation(img_array,
+def segmentation(param,
```
remtav, line 139: here again, docstrings could help understand what arguments are expected. Not sure I understand the tp_mem argument. Does it stand for "temporary memory"? It's the image to be inferred?

inference.py
```diff
@@ -43,112 +45,223 @@
...
+    model.eval()
+
+    # initialize test time augmentation
+    transforms = tta.Compose([tta.HorizontalFlip(), ])
+    # construct window for smoothing
+    WINDOW_SPLINE_2D = _window_2D(window_size=padded, power=2.0)
+    WINDOW_SPLINE_2D = torch.as_tensor(np.moveaxis(WINDOW_SPLINE_2D, 2, 0), ).type(torch.float)
+    WINDOW_SPLINE_2D = WINDOW_SPLINE_2D.to(device)
+
+    fp = np.memmap(tp_mem, dtype='float16', mode='w+', shape=(h_, w_, num_classes))
```
remtav, line 176: can you add more comments from here on? Why do you create a memory map from tp_mem? What does it become further on?

inference.py
```diff
@@ -43,112 +45,223 @@
...
+    subdiv = 2.0
+    step = int(chunk_size / subdiv)
+    pred_img = np.zeros((h_, w_), dtype=np.uint8)
+    for row in tqdm(range(0, input_image.height, step), position=2, leave=False):
+        for col in tqdm(range(0, input_image.width, step), position=3, leave=False):
+            arr1 = fp[row:row + chunk_size, col:col + chunk_size, :] / (2 ** 2)
+            arr1 = arr1.argmax(axis=-1).astype('uint8')
+            pred_img[row:row + chunk_size, col:col + chunk_size] = arr1
+    pred_img = pred_img[:h, :w]
+    end_seg = time.time() - start_seg
```
remtav, line 253: here again, comments would help follow what's going on. Many lines are similar to those in gen_img_sample(). Could this be refactored?

utils/utils.py
```diff
@@ -487,37 +527,6 @@ def ordereddict_eval(str_to_eval: str):
         return str_to_eval
 
 
-def defaults_from_params(params, key=None):
```
remtav, line 490: this was added by Blaise, right? Why remove it?

inference.py
victorlazio109: Yeah, it does not work well with the revised smoothening function. I can explain further. It has to do with the padding of the image x2 of its dimension, so you have to provide a static chunk_size which will not burst the memory during operations.

gdl_hyperopt_template.py
victorlazio109: can be worked on, previous work from an intern

inference.py
victorlazio109: I would add docstrings amd make proper comments. tp_mem is a temp_file to write very large arrays to disk at inference

inference.py
victorlazio109: most probably, code improvement is always work in progress

inference.py
victorlazio109: basically a temporary memory map file was introduced to aid inference on very large tifs

utils/utils.py
victorlazio109: I think this was refactored somewhere else and it became redundant

gdl_hyperopt_template.py
victorlazio109: This will be taken off, it is unique to my special use case of hyperopt.

sample_creation.py
victorlazio109: not necessarily, it exists to support the new way our images(*single bands) are called and processed. Let me know if you have suggestions where you want it to live.

mpelchat04 (approved):

CharlesAuthier (approved):

train_segmentation.py
```diff
@@ -616,16 +609,6 @@ def main(params, config_path):
...
-    # list of GPU devices that are available and unused. If no GPUs, returns empty list
-    gpu_devices_dict = get_device_ids(num_devices,
-                                      max_used_ram_perc=max_used_ram,
-                                      max_used_perc=max_used_perc)
-    logging.info(f'GPUs devices available: {gpu_devices_dict}')
-    num_devices = len(gpu_devices_dict.keys())
-    device = torch.device(f'cuda:{list(gpu_devices_dict.keys())[0]}' if gpu_devices_dict else 'cpu')
-
-    logging.info(f'Creating dataloaders from data in {samples_folder}...\n')
-
```
remtav, line 628: Vic, any reason why you removed this? Were you having trouble with it?

inference.py
```diff
@@ -412,34 +542,37 @@ def main(params: dict):
...
-                inference_image = working_folder.joinpath(local_img.parent.name,
-                                                          f"{img_name.split('.')[0]}_inference.tif")
-            with rasterio.open(local_img, 'r') as raster:
-                logging.info(f'Reading image as array: {raster.name}')
-                img_array, raster, _ = image_reader_as_array(input_image=raster, clip_gpkg=local_gpkg)
-                if debug:
-                    logging.debug(f'Unique values in loaded raster: {np.unique(img_array)}\n'
-                                  f'Shape of raster: {img_array.shape}')
+        for info in tqdm(list_img, desc='Inferring from images', position=0, leave=True):
+            with start_run(run_name=Path(info['tif']).name, nested=True):
```
remtav, line 546: There's a little bug here. at line 475, start_run is imported only if a mlflow uri is inputted. We can just import everything at the top of the script no matter if the uri is given or not. I can address this bugfix in my next PR.

train_segmentation.py
victorlazio109: moved to model_choice.py fixed too with the frequent bug on HPC of not being able to use GPUs with ids other than 0

PR #247, opened 2022-01-26
Dockerfile modifications

### LINKED ISSUE #241: Dockerfile should be optimized to generate a smaller runtime image
Hodgepodge of reasons:

- Don't use a single layer image a.k.a multi-stage images; "always create your own base image layer for your OS, another layer for the username definition, another layer for the runtime installation, another layer for the configuration, and finally another layer for your application¹
-  what matters is the size of the frequently changing layers. And that’s what should be minimized ... layers that are changed most frequently should be created last ... Having a smaller Docker image is generally desirable ... for a number of reasons like reducing the attack surface ... - When a Docker image is uploaded or downloaded (pushed or pulled), each image layer is transferred separately. The layers that already exist at the destination are not transferred at all ... In plain English: the first transfer takes the hit. In the following runs, only the new (frequently changing) layers need to be transferred. And their size depends on the image structure. A somewhat exceptional situation is a CI/CD service. In such setups, computing resources are temporary, so pulled Docker images don’t persist between runs. Because of that, the entire image has to be pulled again and again on each step of the pipeline²
-  Docker containers support the implementation of CI/CD in development. Image size and build efficiency are important factors when overseeing and working with the microservice architecture. This is why you should try to keep your Docker images small³
- Bundle commands in `RUN` commands.  Layers are like git commits. Docker layers store the difference between the previous and the current version of the image. When a Git repository becomes bigger, you can choose to squash the history into a single commit and forget about the past.  It turns out you can do something similar in Docker too with a multi-stage build.⁴

¹ https://developers.redhat.com/blog/2016/02/24/10-things-to-avoid-in-docker-containers
² https://semaphoreci.com/blog/2018/03/14/docker-image-size.html
³ https://phoenixnap.com/kb/docker-image-size
⁴ https://learnk8s.io/blog/smaller-docker-images

ymoisan:
Multi-stage build allows bringing image size down about 30% to ~ 8.6 GB.  Still work to do to really close #241


remtav:
Thanks! Can you delete the [docker compose yaml](https://github.com/ymoisan/geo-deep-learning/blob/docker/docker-compose.yml)? This PR is a good occasion to do so as it's not relevant and not sure how and why it got there...

Dockerfile
```diff
@@ -1,18 +1,82 @@
...
+## Variables 
+ARG GIT_HASH=f7e6033
+
+RUN git clone "https://github.com/NRCan/geo-deep-learning.git" && \
+    cd geo-deep-learning && \
+    git checkout ${GIT_HASH}
+
+# TODO : create an environment for inference
+RUN conda env create -f geo-deep-learning/environment.yml && \
+    rm -rf .git data docs config example LICENSE README.md Dockerfile *.yml
```
remtav, line 34: - the "config" folder is necessary and should not be removed in my opinion. 
- "example" folder contains an empty **22 bytes** README to be populated with an example usage in a notebook
- the "data" folder is necessary for a user to test GDL with the Massachusetts dataset locally. 

I don't think this rm -rf command is relevant other than for deleting .git history, which doesn't need to be important in the first place if we use git tags.

remtav (changes requested):

Dockerfile
```diff
@@ -1,18 +1,82 @@
...
+## Variables 
+ARG GIT_HASH=f7e6033
+
+RUN git clone "https://github.com/NRCan/geo-deep-learning.git" && \
+    cd geo-deep-learning && \
+    git checkout ${GIT_HASH}
+
+# TODO : create an environment for inference
+RUN conda env create -f geo-deep-learning/environment.yml && \
+    rm -rf .git data docs config example Dockerfile *.yml
```
remtav, line 34: I don't think we should delete all these folders. See [my comment](https://github.com/NRCan/geo-deep-learning/pull/247#discussion_r793703835)

remtav (changes requested):

Dockerfile
```diff
@@ -1,18 +1,82 @@
...
+    rm -rf /tmp/* && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
 
-# The code to run when container is started:
-#COPY run.py .
-ENTRYPOINT ["conda", "run", "--no-capture-output", "-n", "geo_deep_env", "python"]
\ No newline at end of file
+## Variables 
+ARG GIT_HASH=f7e6033
```
remtav, line 26: This hash refers to an old state of gdl. latest version on develop is:
version: v2.0.1
hash: 23abf68

Dockerfile
```diff
@@ -1,18 +1,82 @@
-FROM continuumio/miniconda3
+FROM nvidia/cuda:11.2.0-cudnn8-devel-ubuntu20.04 AS build
```
remtav, line 1: See what [RasterVision has done](https://github.com/azavea/raster-vision/blob/master/Dockerfile#L1), maybe the cuda version could be inputted by the user as an variable.

Dockerfile
```diff
@@ -1,18 +1,82 @@
...
+# Create the user
+RUN useradd --create-home -s /bin/bash --no-user-group -u $USERID $USERNAME && \
+    chown $USERNAME $CONDA_DIR -R && \
+    adduser $USERNAME sudo && \
+    echo "$USERNAME ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
+USER $USERNAME
+WORKDIR /home/$USERNAME
+
+## Variables 
+ARG GIT_HASH=f7e6033
```
remtav, line 63: This is a duplicate of line 26

Dockerfile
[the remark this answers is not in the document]
ymoisan: Since the type of tag you are referring to is "[lightweight](https://git-scm.com/book/en/v2/Git-Basics-Tagging)", it’s just a pointer to a specific commit hash.  For deployment purposes we certainly want to build images with a (version) tag.  For development purposes however I think it makes sense to build images on specific commits/hashes.  Ideally we would be able to supply build args of either `hash` or `tag` type.  

I asked a question a few weeks ago that is akin to what I'd like to be able to do here.  In [that case](https://stackoverflow.com/questions/70840481/corporate-ca-certificate-in-git-repo) it was about optionally including a corporate certificate at build time.  I have not found satisfactory answers yet.  I could modify the PR to accept tags instead of hashes but if we'll want to use the Dockerfile to build containers dynamically on committed pushes it does not make sense IMO.

Dockerfile
ymoisan: [Current ](https://github.com/ymoisan/geo-deep-learning/commit/0284acee3919d3a5bca830f96c2ff48939a122c7) does just that.

Dockerfile
ymoisan: > ... {CUDA_VERSION}-cudnn7 ...

This is probably specific to their set up.  I'm not sure it's wise to allow the CUDA_VERSION be passed as an argument.  What if we input 11.2 and there were no cudnn7/8 Ubuntu images for that CUDA version ?

Also Nvidia has a [CUDA compatibility matrix](https://docs.nvidia.com/deeplearning/cudnn/support-matrix/index.html) which shows that CUDA versions have to go along with the GPU host's CUDA Driver version.  I'd rather stick with CUDA and cudnn versions that we know works on our systems.

Dockerfile
ymoisan: No : ARGs only last for the build phase of a single image. {For the multistage, renew the ARG](https://stackoverflow.com/questions/53681522/share-variable-in-multi-stage-dockerfile-arg-before-from-not-substituted).  I did not know I didn't have to specify the `ARG` value in a subsequent stage though.  Will fix.

Dockerfile
ymoisan: If we use tags then what happens with CI if we want to dynamically build the imeage ?

Dockerfile
ymoisan: `GIT_HASH` is a build arg.  We can edit the Dockerfile with a new `GIT_HAS` periodically (e.g. when tagging a new version) but in the general case we'll use whatever commit hash at build time e.g. for CI.

```Dockerfile
docker build --build-arg GIT_HASH=23abf68 ...
```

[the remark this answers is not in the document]
remtav: I understand you are looking at the maximum flexibility possible when cloning, but I still disagree with the use of hash. Git hashing is useful as an "internal unique identifier" for git, but I don't think it's meant to be used the way we use it here. I see two main disadvantages:
- Whether it's tags of hashes the user will still have to manually change the Dockerfile to get the version of GDL he/she wants (a "latest" tag like in Dockerhub could be cool). If using hashes, this means one needs to get the desired hash everytime one wants to perform a git clone. Since git hashes don't "mean anything" in terms of GDL's state (i.e. hashes don't give you a hint on the year/month, PR or anything else you are about to clone), it seems like pointing to a specific tag is more "meaningful". For the purpose of using the Dockerfile successfully, the presupposes that the developers of GDL push a tag **at least** when dependencies change (which happens no more than a few times a year). Otherwise, we could require devs to push a tag at every bugfix or so (2.0.1 -> 2.0.2).
- The other disadvantage is in the Docker build flow itself. By using a tag, and therefore making it possible to git clone with the [--depth ](https://www.git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthltdepthgt) parameter, the git history is automatically discarded. If using hash, the git history is downloaded (about 100 Mb), then deleted. This is not optimal and extends the build by a few unnecessary  seconds IMO.

Dockerfile
remtav: See [previous comment](https://github.com/NRCan/geo-deep-learning/pull/247#discussion_r797670749). CI will only break if dependencies change. When dependencies change, a new version should be tagged as good practice. Therefore the Dockerfile should be changed along with the tag.

Dockerfile
remtav: Indeed, I ran into the "What if we input 11.2 and there were no cudnn7/8 Ubuntu images for that CUDA version" problem. It is a big downside.

remtav (approved):

Dockerfile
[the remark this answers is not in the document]
ymoisan: > Whether it's tags of hashes the user will still have to manually change the Dockerfile to get the version of GDL he/she want

No.  The HASH is a build argument : `--build-arg GIT_HASH=your-hash`.  We can edit the Dockerfile periodically to put a "good default" but other than that it is totally overridable at build time.  That way the Dockerfile is parametrized, so when doing CI for instance you could dynamically build an image with the latest commit for tests.  If only we could specify either HASH or TAG, that would make the Dockerfile very agile.

For now, I suggest you test the resulting image and make sure it works as expected.  If it does, then we'll have done "iteration 1" of Dockerfile optimization and I think we could merge this PR.  Note : The image now weighs 9.5GB for a reason I can't understand ...

About that sudden extra GB that came about with this last version of Dockerfile, I would open another issue with the idea of specializing Dockerfiles : one for CI (probably a one-stage who-cares-how-fat-it-is), one for publication (e.g. on DockerHub with git TAGS) and one for inference, as it does not require all the dependencies we pack in our Docker image.  How does that sound ?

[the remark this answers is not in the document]
remtav: Ok for setting HASH as build argument, but still, 

> this means one needs to get the desired hash everytime one wants to perform a git clone

. The disadvantage I pointed out is still there.

I approved the PR for the sake of moving forwards beyond this debate, but I'm still far from convinced by the use git hash here.

victorlazio109 (approved):

Issue #246, opened 2022-01-25
Model choice: maximize Hydra's features


remtav:
Currently, choosing a model in GDL with [model_choice.py](https://github.com/NRCan/geo-deep-learning/blob/develop/models/model_choice.py) makes use of a series of disorganized if-else statements.

As mentionned in this [Towards Data Science article](https://towardsdatascience.com/complete-tutorial-on-how-to-use-hydra-in-machine-learning-projects-1c00efcc5b9b), choosing models based on a series of if-else conditions is to be avoided. Now that hydra is implemented with PR #208, choosing a model can be much more straight forward by using config files for each relevant architecture and pointing to the model class directly in this config file. 

This was already done by @CharlesAuthier with [fastrcnn](https://github.com/NRCan/geo-deep-learning/blob/develop/config/model/fastrcnn.yaml).

PR #256, opened 2022-02-01
use hydra to instantiate loss

remtav:


config/loss/multiclass/ohemcrossentropy.yaml
```diff
@@ -0,0 +1,4 @@
+# @package _global_
```
remtav, line 1: Should we name this file ohemce.yaml or keep full name?

config/training/default_training.yaml
```diff
@@ -8,7 +8,6 @@ training:
   max_epochs: ${general.max_epochs}
   min_epochs: ${general.min_epochs}
   num_workers:
-  loss_fn: CrossEntropy  # name of the metric you want to manage the training
```
remtav, line 11: Ideally, the loss parameter should stay in the training config, but I couldn't point this parameter to the one in config/loss/multiclass use hydra's ${...} syntax. I did try quite a bit. I'd be moving the loss parameter out of training if we can't find a way to make this work.

models/model_choice.py
```diff
@@ -137,9 +138,10 @@ def set_hyperparameters(params,
     gamma = get_key_def('gamma', params['scheduler']['params'], 0.9)
     class_weights = torch.tensor(class_weights) if class_weights else None
     # Loss function
-    criterion = MultiClassCriterion(loss_type=loss_fn,
-                                    ignore_index=dontcare_val,
-                                    weight=class_weights)
+    if loss_fn['_target_'] == 'torch.nn.CrossEntropyLoss':
+        criterion = instantiate(loss_fn, weight=class_weights)  # FIXME: unable to pass this through hydra
```
remtav, line 142: I really don't like this. CrossEntropy loss (multiclass) is the only loss that uses class_weights as a parameter. We had adapted all other losses to accept it but no use it. The binary losses that @victorlazio109 has added from smp don't accept the "weight" parameter and therefore raise an exception if this parameter is passed.

Logically, class_weights would have been in the crossentropy.yaml config directly (and not in model_choice). It could therefore be dealt with directly by hydra, but one things makes it a bit tricky:
Cross entropy in torch.nn [only accepts a tensor](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/loss.py#L1095). If it could handle a list, this would be much more convenient for hydra.

The only way hydra can directly input a tensor to the weight argument is by changing the CE yaml to somtehing like this:
# @package _global_
loss:
```
    _target_: torch.nn.CrossEntropyLoss
    ignore_index: ${dataset.ignore_index}
    weight:
        _target_: torch.tensor
        data: ${dataset.class_weights}
```
This works fine when class_weights is not None (ex.: [0.5,0.5] successfully gets converted to tensor), but when it's None torch.tensor(None) raises and exception. It would require hydra to have some king of if/else statement if class_weights is None.

Anyways, I'm leaving it to you guys if you have ideas. I spent enough time on this case.

train_segmentation.py
```diff
@@ -325,6 +325,12 @@ def training(train_loader,
                                ep_num=ep_idx + 1,
                                scale=scale)
 
+        # FIXME: binary vs multiclass loss don't expect same dimensions
+        try:
+            if criterion.mode == 'binary':
+                outputs = outputs[:, -1, ...]
```
remtav, line 331: @victorlazio109 did you run into this "dimensionality" problem when using binary losses? When testing out, I noticed the binary losses only accept logits without the "backgound". Therefore, the prediction array must be indexed as shown in this bit of ugly code from [1,2,256,256] (each number being [batch size, num_classes, width, height]) to [1, 256, 256], discarding the background array and therefore deflating the dimensions to match those of the labels [1, 256, 256]

config/loss/binary/dice.yaml
```diff
@@ -0,0 +1,5 @@
+# @package _global_
+loss:
+    _target_: segmentation_models_pytorch.losses.DiceLoss
+    mode: binary
+    ignore_index: ${dataset.ignore_index}
```
remtav, line 5: I don't think these missing newlines will cause trouble. I can add them if anybody's nervous :)

train_segmentation.py
```diff
@@ -519,7 +519,7 @@ def train(cfg: DictConfig) -> None:
 
     # MODEL PARAMETERS
     class_weights = get_key_def('class_weights', cfg['dataset'], default=None)
-    loss_fn = get_key_def('loss_fn', cfg['training'], default='CrossEntropy')
+    loss_fn = cfg.loss
```
remtav, line 522: the get_key_def can't be used as is since the loss will later be instantiated with hydra. The loss parameter is not just a string that can default to "crossentropy"

CharlesAuthier (dismissed):

train_segmentation.py
victorlazio109: resolved with my last PR, confirm if you can

config/loss/binary/dice.yaml
victorlazio109: where is it passed if removed ?

models/model_choice.py
```diff
@@ -136,12 +135,11 @@ def set_hyperparameters(params,
     gamma = get_key_def('gamma', params['scheduler']['params'], 0.9)
     class_weights = torch.tensor(class_weights) if class_weights else None
     # Loss function
-    if num_classes == 1:
-        criterion = SingleClassCriterion(loss_type=loss_fn, ignore_index=dontcare_val)
+    if loss_fn['_target_'] in ['torch.nn.CrossEntropyLoss', 'losses.focal_loss.FocalLoss',
```
victorlazio109, line 138: We lose the differentiator between SingleClass and MultiClass here, if a user were to pass num_classes == 1 while picking CE loss he will almost certainly run into errors at some point in the code execution

config/loss/binary/dice.yaml
remtav: The only script that is removed is losses/__init__.py. It is replaced by a colleciton of yaml, one for each loss. Each yaml points to the class that needs to be instantiated by Hydra. It's the "hydranic" way of things that will also appear once we tackle issue #246

models/model_choice.py
remtav: Good point. I can add a small check for this.

config/loss/multiclass/ohemcrossentropy.yaml
CharlesAuthier: I agree less long

train_segmentation.py
```diff
@@ -519,7 +514,10 @@ def train(cfg: DictConfig) -> None:
...
+    if not loss_fn.is_binary == (num_classes == 1):
+        raise ValueError(f"A binary loss was chosen for a multiclass task")
+    del loss_fn.is_binary  # prevent exception at instantiation
```
remtav, line 520: @victorlazio109 this is the simplest way I've found to make sure binary losses are not called in multiclass problem and vice versa.

victorlazio109 (approved):

PR #262, opened 2022-02-03
Reimplement classification task: oups!

remtav:
Sorry, I didn't know/remember that we completely abandonned the classification task. Since I saw bits of it in the inference script, I though it was just on the TODO list... My main motivation was to clean the inference_segmentation.py script, not reimplement classification. Oups.

Featuring:
resuscitate train_classification.py
resuscitate inference_classification.py
add classif template config
replace load_obj() with hydra's get_method
remove task config directory


remtav:
We have decided to definitely remove classification task from GDL.

Issue #264, opened 2022-02-03
CI: shrink inference data


remtav:
[Currents CI testing](https://github.com/NRCan/geo-deep-learning/runs/5058568329?check_suite_focus=true) take 15 minutes, but 11 minutes are dedicated to inference on 2 massachusetts' dataset images. 

This is outrageously long for CI tests considering it doesn't add any value to the testing: we know after a minute or so that inference works.

CharlesAuthier:
good idea maybe change the number of images for inference like a small 512 by 512

Issue #243, opened 2022-01-25
Reimplement hyperparameter optimization with hyperopt, optuna or other


remtav:
This feature from PR #201 was broken with introduction of Hydra in PR #208. Needs to be reimplemented.

remtav:
With Hydra, using the [Optuna plugin ](https://hydra.cc/docs/next/plugins/optuna_sweeper/) seems a better choice for hyperparameter optimization than hyperopt. The plugins leverages Hydra's [multirun feature](https://hydra.cc/docs/1.0/tutorials/basic/running_your_app/multi-run/).

remtav:
Will tackle #152 first, then cover this specific feature if needed.

Issue #138, opened 2020-05-26
Update data_analysis.py according to last devs on gdl


remtav:
For example, parameters in yaml should be identical for data_analisys and for samples preparation with images_to_samples.py.

By reformatting those parameters in yaml, the way data_analisys.py deals with the reshaped inputs must be addressed.

Other needs for developping this very useful module could also arise.

PR #267, opened 2022-02-08
Refactor inference and apply refactoring to other scripts

remtav:


.github/workflows/github-actions-ci.yml
```diff
@@ -23,7 +23,6 @@ jobs:
         run: |
           source /usr/share/miniconda/etc/profile.d/conda.sh
           conda activate geo_deep_env
-          unzip ./data/massachusetts_buildings.zip -d ./data
```
CharlesAuthier, line 26: are we not testing on that dataset anymore?

README.md
```diff
@@ -57,12 +57,7 @@ git clone https://github.com/NRCan/geo-deep-learning.git
 cd geo-deep-learning
 ```
 
-2. Unzip the _**massachusetts buildings**_ dataset.
-```shell
-unzip ./data/massachusetts_buildings.zip -d ./data
-```
-
-3. Run the wanted script (for segmentation).
+2. Run the wanted script (for segmentation).
```
CharlesAuthier, line 60: That was for an running example  out of the box

gdl_hyperopt_template.py
```diff
@@ -17,8 +17,8 @@
 import torch
 # ToDo: Add hyperopt to GDL requirements
 from hyperopt import fmin, tpe, hp, Trials, STATUS_OK
+from ruamel_yaml import YAML
```
CharlesAuthier, line 20: you have remove it form the env yaml in this same PR

environment.yml
```diff
@@ -20,7 +20,6 @@ dependencies:
   - python>=3.9.2
   - pytorch=1.10.0=py3.9_cuda10.2_cudnn7.6.5_0
   - pyyaml>=5.4.1
-  - ruamel_yaml>=0.15.80
```
CharlesAuthier, line 23: You remove it but call it in other files

train_classification.py
```diff
@@ -0,0 +1,435 @@
+import time
```
CharlesAuthier, line 1: Why do you had back the training classification?

CharlesAuthier (changes requested):

.github/workflows/github-actions-ci.yml
remtav: This PR was built on top of PR #264 . This modification belongs to the other PR. The test data is simply unzipped by default. I've started working on #257 and all test data will be this way (ie saved as is, not zipped)

README.md
remtav: See previous comment. The test data won't be zipped anymore.

environment.yml
remtav: It is called in deprecated files like gdl_hyperopt_template.py and data_analysis.py. These files are either waiting to be reimplemented with hydra (#243) (ie without ruamel_yaml) or we haven't decided if we keep them or not (#138).

train_classification.py
remtav: oups thanks for pointing out. Got mixed up in my rebasing.

remtav:
I've decided to split this PR into smaller ones to focus on specific changes for each PR. It will be more in line with our workflow.

Issue #223, opened 2021-12-15
Reading single-band rasters: unified approach with multiband reader


remtav:
As operationalization approaches, inference.py will need to read and concatenate single band rasters rather than [multiband only](https://github.com/NRCan/geo-deep-learning/blob/develop/inference.py#L120). This implies extending [sample_creation.py](https://github.com/NRCan/geo-deep-learning/blob/develop/sample_creation.py)'s use to inference.py and possibly merging with [images_to_samples.py](https://github.com/NRCan/geo-deep-learning/blob/develop/images_to_samples.py) (soon to become data_to_tiles.py --> #215 ). 

This issue is closely related to #221 and #222. A future reading utility will have to manage a list of rasters as input with variable configration (some may be multiband, others will be single-band, maybe even for a same training dataset). This utility will then be available for tiling as well as inference.

ymoisan:
With single-band COGs, it may be possible to use dynamic capabilities (e.g. NRCan's datacube) to create streamable (i.e. cloud optimized) multiband images ready for training or inference.  GDL would then be presented with ready-to-use N-band images for sample creation/tiling.

This ticket is about N bands of one image.  The wider use case of N-band fused data (e.g. lidar resampled to an image gsd and used in conjunction with it for training) may be better addressed with a ready-to-use N-band image generation pre-processing step.

remtav:
This issue is becoming a priority to address as the temporary [sample_creation.py](https://github.com/NRCan/geo-deep-learning/blob/develop/sample_creation.py) for tiling single-band imagery is getting more and more outdated.

Here's what I'd suggest to make a uniform tiling pipeline for both single-band ([stac item](https://github.com/NRCan/geo-deep-learning/issues/222) or not) and multi-band imagery. In the [proposed input csv ](https://github.com/NRCan/geo-deep-learning/issues/221#issuecomment-1121310523), the imagery path could be in one of three formats:
1. path to a multi-band image file:
`my_dir/my_multiband_geofile.tif`

2. path to a single-band image file, where the band specific string in the file name would be in a hydra-like [interpolation format](https://hydra.cc/docs/1.0/advanced/override_grammar/basic/#primitives). The interpolation string would refer to a dataset parameter with a list of desired band names or strings to help resolve the single-band filenames. For example:
in dataset config:
`dataset.bands: [R, G, B]`
in input csv:
`my_dir/my_singleband_geofile_band_${dataset.bands}.tif`
[...]

Issue #222, opened 2021-12-15
Reading training and inference rasters from STAC item list: input as json


remtav:
Soon, GDL will internally be used to read images from an S3 bucket, using STAC items and Franklin's API to submit requests based on spatial, temporal or other filters from metadata. 

GDL's readers should therefore transition from csv as input (#221) to a json containing a list of STAC items (as json). Here's an example of a response from a request passed to Franklin that could directly be used by GDL in the future:
`{
    "type": "FeatureCollection",
    "context":
    {
        "returned": 1,
        "matched": 1
    },
    "features":
    [
        {
            "id": "QC15_052686187030_01_QuickBird2",
            "stac_version": "1.0.0-beta.2",
            "stac_extensions":
            [
                "https://stac-extensions.github.io/projection/v1.0.0/schema.json"
            ],
            "type": "Feature",
            "geometry":
            {
                "type": "Polygon",
                "coordinates":
                ...
    ],
    [ (2nd STAC item)
    ],
    [ (3rd STAC item)
    ],
    "links":
    [],
    "stac_version": "1.0.0"
}`

Here are 5 examples of STAC Items for IA-Ready images as served by Franklin's API:
https://datacube-stage.services.geo.ca/api/collections/worldview-3-ortho-pansharp/items/QC22_055128844110_01_WV3
https://datacube-stage.services.geo.ca/api/collections/worldview-4-ortho-pansharp/items/MB8_059380728010_01_P001_WV4
https://datacube-stage.services.geo.ca/api/collections/worldview-2-ortho-pansharp/items/VancouverP003_054230029070_01_P003_WV2
https://datacube-stage.services.geo.ca/api/collections/geoeye-1-ortho-pansharp/items/AB13_059294491080_01_GeoEye1
https://datacube-stage.services.geo.ca/api/collections/quickbird-2-ortho-pansharp/items/QC15_052686187030_01_QuickBird2

An issue remains: how to link the ground truth to this list? A separate json? This needs further analysis.

ymoisan:
> how to link the ground truth to this list?

There may be an answer in the [STAC Label](https://github.com/stac-extensions/label) extension.

PR #208, opened 2021-09-16
switch to hydra

CharlesAuthier:
New features introduced by the Hydra structure:
- monitor the logs generated by the code and save it inside a file.
- new documentations on this new structure.
- a config file structured by sections and compartiments to easily find the information wanted or to add new features to GDL.
- a new way to launch task using a main code to redirect the code to the wanted task.
- introduce the early stopping option on plateau

Removed features (to be reimplemented if desired):
- In sampling mode, choose a subset of classes from ground truth based on values for given attribute in gpkg (see [parameter "target_ids"](https://github.com/NRCan/geo-deep-learning/pull/208/files#diff-c0a91f559cabbd9dcf39d1e61e25ad99e6d13a4c8196c35594449313db8fefb5R453))
- In inference mode, if state_dict provided from command line, set architecture accordingly (i.e. overwrite default architecture). For example, the following command doesn't work yet:
`python GDL.py mode=inference inference.state_dict_path="/path/to/custom_checkpoint.pth.tar"`
- use of hyperopt
- use of [metamap](https://github.com/NRCan/geo-deep-learning/pull/208/files#diff-c0a91f559cabbd9dcf39d1e61e25ad99e6d13a4c8196c35594449313db8fefb5R439) and [coordconv](https://github.com/NRCan/geo-deep-learning/pull/208/files#diff-a48bb4c26adf0e214debf7f59dfe46a549b643ab736898d5146afed32c103e48R613) features from PR #82

PR #270, opened 2022-02-09
Housecleaning

### LINKED ISSUE #265: Inference: refactoring needed
In preparation of issues #222 and #223, some refactoring is needed:
- use get_key_def to convert to pathlib.Path objects and to check existence of files/directories
- restore state_dict input to usage before #208 
- tracker: create set_tracker function to remove block of lines about tracker in inference_segmentation.py
- remove unused functions
- a few more minor things.

remtav:
- remove unused functions
- remove ruamel_yaml import from active scripts
- fix dontcare2background related to PR #256

partially closes #265


gdl_hyperopt_template.py
```diff
@@ -13,12 +13,12 @@
 import pprint
```
remtav, line 13: Should we remove this script and get it back in old versions of GDL when ready to reimplement? #243 
Not sure it's relevant to keep it and maintain it for now.

sample_creation.py
```diff
@@ -15,17 +15,29 @@
 from tqdm import tqdm
```
remtav, line 15: Should we remove this script and get it back in old versions of GDL when ready to reimplement? #223 
Not sure it's relevant to keep it and maintain it for now.

utils/data_analysis.py
```diff
@@ -2,11 +2,10 @@
 import logging
```
remtav, line 2: Should we remove this script and get it back in old versions of GDL when ready to reimplement? #138 
Not sure it's relevant to keep it and maintain it for now.

sample_creation.py
CharlesAuthier: What was use for ?

utils/data_analysis.py
CharlesAuthier: Yeah we should remove it

utils/logger.py
```diff
@@ -54,6 +55,12 @@ def save_logs_to_bucket(bucket, bucket_output_path, output_path, now, batch_metr
 
 
 def dict2path(my_dict, path=None):
+    """
+    TODO
+    @param my_dict:
+    @param path:
+    @return:
+    """
```
CharlesAuthier, line 63: We should probably fill those?

utils/logger.py
```diff
@@ -66,7 +73,27 @@ def dict2path(my_dict, path=None):
 
 
 def dict_path(param_dict, param_name):
+    """
+    TODO
+    @param param_dict:
+    @param param_name:
+    @return:
+    """
```
CharlesAuthier, line 81: Same comment

CharlesAuthier (dismissed):

sample_creation.py
remtav: It was used to received single-band imagery as input. This will be tackled in the coming month or so with issue #223 and #222 . I think we can remove it if @victorlazio109 is okay with that. Otherwise, we can wait and remove it when a different solution is merged.

gdl_hyperopt_template.py
remtav: @victorlazio109 is that okay with you? With hydra, it may be cool to go with optuna for hyperparameter search #243

utils/data_analysis.py
remtav: okay with you @mpelchat04 and @victorlazio109 ? We can always resuscitate/reimplement it if necessary

utils/logger.py
remtav: I agree! I added a TODO to make sure we don't forget. I think you were the one to add them at PR #208 . Do you think you can add docstrings for these functions?

gdl_hyperopt_template.py
remtav: EDIT: No need to remove gdl_hyperopt_template. It will be removed once #278 is merged.

mpelchat04 (approved):

PR #272, opened 2022-02-09
Manage device choice at a single place with set_device

### LINKED ISSUE #269: Refactor: manage device selection and error-handling (HPC) with set_device function
The management of training/inference device (cuda vs cpu) is currently cumbersome in the code. Some example:
https://github.com/NRCan/geo-deep-learning/blob/6b037e12c099731e913cff72d5bd6a6cecd9d208/models/model_choice.py#L256
https://github.com/NRCan/geo-deep-learning/blob/6b037e12c099731e913cff72d5bd6a6cecd9d208/train_segmentation.py#L230

Also there a multiple loggings to indicate which device is selected:
https://github.com/NRCan/geo-deep-learning/blob/6b037e12c099731e913cff72d5bd6a6cecd9d208/models/model_choice.py#L264
https://github.com/NRCan/geo-deep-learning/blob/6b037e12c099731e913cff72d5bd6a6cecd9d208/utils/utils.py#L128

A set_device function could manage this aspect in a unique place and some logging calls could be removed.

remtav:
closes #269


utils/logger.py
```diff
@@ -66,7 +73,27 @@ def dict2path(my_dict, path=None):
...
     }
+
+
+def get_logger(name=__name__, level=logging.INFO) -> logging.Logger:
+    """Initializes multi-GPU-friendly python logger."""
+
+    logger = logging.getLogger(name)
+    logger.setLevel(level)
+
+    # this ensures all logging levels get marked with the rank zero decorator
```
victorlazio109, line 94: cool!

utils/utils.py
```diff
@@ -184,6 +165,26 @@ def gpu_stats(device=0):
...
+    @param gpu_devices_dict: dictionary containing info on GPU devices as returned by lst_device_ids
+    @return: torch.device
+    """
+    if gpu_devices_dict:
+        logging.info(f"\nCuda devices available: {gpu_devices_dict}.\nUsing {list(gpu_devices_dict.keys())[0]}\n\n")
+        device = torch.device(f'cuda:{list(range(len(gpu_devices_dict.keys())))[0]}')
+    else:
+        logging.warning(f"\nNo Cuda device available. This process will only run on CPU")
+        device = torch.device('cpu')
+        try:
```
victorlazio109, line 180: I don't think this try statement necessary

utils/utils.py
```diff
@@ -184,6 +165,26 @@ def gpu_stats(device=0):
...
+            models.resnet18().to(device)  # test with a small model
+        except (RuntimeError, AssertionError):  # HPC: when device 0 not available. Error: Cuda invalid device ordinal.
```
victorlazio109, line 182: seems the comment there is no longer valid there, over the summer we solved the problem of Cuda invalid device ordinal. We can just say # Error !

inference_segmentation.py
```diff
@@ -401,6 +403,11 @@ def main(params: dict) -> None:
         logging.warning(f'\nDebug mode activated. Some debug features may mobilize extra disk space and '
                         f'cause delays in execution.')
 
+    # list of GPU devices that are available and unused. If no GPUs, returns empty dict
+    gpu_devices_dict = get_device_ids(num_devices, max_used_ram_perc=max_used_ram, max_used_perc=max_used_perc)
+    chunk_size = get_key_def('chunk_size', params['inference'], default=512, expected_type=int)
+    chunk_size = calc_inference_chunk_size(gpu_devices_dict=gpu_devices_dict, max_pix_per_mb_gpu=50, default=chunk_size)
```
victorlazio109, line 409: just a note: calculating chunk size dynamically here specifically at inference may cause problems with test time augmentation and smoothing. A default size of 512 is advised. 512 is padded to 512 * 2 during test time routines

utils/utils.py
remtav: Ok! If you say the problem is solved, I can simply remove the try/except all together. Thanks!

inference_segmentation.py
remtav: ok, do you agree to keep the "automatic chunk_size" option? it's meant to be tunable, i.e. max_pix_per_mb_gpu can be lowered if the automatic way busts memory.

utils/logger.py
remtav: this was @CharlesAuthier addition though :)

inference_segmentation.py
mpelchat04: As discussed we can lower the  `max_pix_per_mb_gpu` for now, keeping in mind that we could later optimize its definition during inference.

CharlesAuthier (approved):

mpelchat04 (approved):

inference_segmentation.py
remtav: True. I'll address this in PR #276

Issue #215, opened 2021-12-07
Images_to_samples.py (hdf5s) --> data_to_tiles.py (tiles as tifs and geojsons)


remtav:
# Features
- CRS mismatch automatic handling
- Resize image to new spatial resolution (ex.: 50cm to 25cm)
- Don't rewrite tiles for an image that has all expected tiles
- Crop geopackage to geojson (don't burn right away)
- Input a csv directly (rush mode)
- Input an image directly (for inference if pre-tiling is preferred?)
- Parallelize tiling

# Where to write tiles?
[data_path / experiment_name / samples_folder_name](https://github.com/remtav/geo-deep-learning/commit/83d70d4c41fc225abc3d0fd649e36ba63488f07f#diff-614a3bfd3eccc7c4b81cb8d8fc8ad1c68c5198e1b568bfa9813f3ad1e24e8045R205) ?
ex.: 
`geo_deep_learning/data/geobase_buildings/tiles_3bands_minannot1`

# How to deal with CRS mismatch?
- to be discussed...

# Onboarding solaris
- How to integrate necessary solaris' utils?
- [1st attempt](https://github.com/remtav/geo-deep-learning/commit/678bb81f9bb9ff3f1e4e2ff888b3f08fdd1a4e34)

remtav:
Suggested logic for data_to_tiles.py

![data_to_tiles_flow drawio](https://user-images.githubusercontent.com/34774759/150404676-37a11af0-5a0b-4da0-80cc-9c5d5030c333.png)

remtav:
Suggested class structure for use in data_to_tiles.py

![classes_data_to_tile drawio](https://user-images.githubusercontent.com/34774759/150404999-22100733-1101-43de-80d1-d4abbc6904c3.png)

remtav:
Seeing comments from team members on Slack and Github, seems like this issue needs further discussion. 
I suggest starting from a pros/cons table, then discussing it in a meeting based on this table.

|Tiling approach|Pros|Cons|
|---|---|---|
|HDF5|*||
|Plain .geotiff/.geojson|||
>*This type of table that expects a list of items is much easier done with Google Doc than [Markdown](https://stackoverflow.com/questions/19950648/how-to-write-lists-inside-a-markdown-table)

[...]

PR #275, opened 2022-02-09
Requirements: add torchgeo, pystac, solaris (v0.5.0), rio-stac

remtav:
necessary to start working on #222 and #215 

In files changed, see [environment.yaml](https://github.com/NRCan/geo-deep-learning/pull/275/files#diff-9efd195f4e9bfb79ccd456a1d8370fafcc4bcb0b00ea3799222667d2ae818533) only.

add dependency: torchgeo, pystac, solaris (v0.5.0), rio-stac
move pip dependencies to conda if possible
change channel priority (pytorch, then conda-forge)
remove secondary dependencies when possible


environment.yml
```diff
@@ -1,60 +1,38 @@
...
-  - defaults
+  - conda-forge
 dependencies:
-  - fiona>=1.8.0
-  - gdal>=3.0.0
   - boto3>=1.19.2
-  - cudatoolkit=10.2.89
+  - cudatoolkit>=11.3.0
+  - pytorch=1.10.1=py3.9_cuda11.3_cudnn8.2.0_0  # Linux
+  # - pytorch=1.10.1=py3.9_cuda11.3_cudnn8_0  # Windows
```
remtav, line 9: Couldn't find a way to make it work without hardcoding this. Currently, things are also harcoded to Linux (see l21 of old version) so this PR is at least not making things worst...

environment.yml
```diff
@@ -1,60 +1,38 @@
...
   - pytorch
-  - defaults
+  - conda-forge
 dependencies:
-  - fiona>=1.8.0
-  - gdal>=3.0.0
   - boto3>=1.19.2
-  - cudatoolkit=10.2.89
+  - cudatoolkit>=11.3.0
+  - pytorch=1.10.1=py3.9_cuda11.3_cudnn8.2.0_0  # Linux
```
ymoisan, line 8: Why not pytorch-1.10.2-py3.9_cuda11.3_cudnn8.2.0_0.  I think you have a typo a=b=c ?

remtav: Just a sheep approach: 1.10.1 has 27k download from conda-forge, where as 1.10.2 has around 8k. We can switch to 1.10.2, I don't see any problem.

environment.yml
ymoisan: I think it's OK.  The version of PyTorch/cuda/cudnn won't change very often.

Your [package match specification](https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#package-match-specifications) is the "three-part" type (two equal signs in the string).  Documentation says:
The third part is always the exact build string. When there are 3 parts, **the second part must be the exact version**. (bold mine)

Tests this morning:
- 1 part `pytorch` ; fetches `pytorch-1.10.2|py3.9_cuda11.3_cudnn8.2.0_0`
- 2 parts; `pytorch=1.10.*1*`; fetches `pytorch-1.10.1|py3.9_cuda11.3_cudnn8.2.0_0 1.21 GB`
- 2 parts; `pytorch=1.10.*2*`; fetches `pytorch-1.10.2             |py3.9_cuda11.3_cudnn8.2.0_0 1.21 GB`

What strikes me here is how the conda package looks like it is the exact same irrespective of PyTorch version.  Since we don't want the PyTorch version to move under our feet, the 1 part package match spec is out.  I think we could use 2 parts at the current latest stable version of PyTorch, 1.10.2.

environment.yml
ymoisan: The newer release gets fewer downloads. Makes sense :-). I think we should stick with the latest stable version. And stick to it for a while unless some major vulnerability comes out.

remtav: Changes made.

CharlesAuthier (dismissed):

remtav:
see #307

PR #278, opened 2022-02-11
Implement optuna hyperparameter optimization

### LINKED ISSUE #243: Reimplement hyperparameter optimization with hyperopt, optuna or other

remtav:
…on (add returns to train_segmentation.py and run_gdl())

metrics.py: add try/except if missing key
model_choice.py: temporarily raise gpu threshold to 100% until issue #246 is addressed
train_segmentation.py:
- fix bug for dontcare value in metrics
- always calculate metrics at test time
inference_segmentation.py: warn if inference has only background values
test hyperparameter optimization in CI with github actions

closes #243


models/model_choice.py
```diff
@@ -244,7 +244,7 @@ def net(model_name: str,
         else:
             checkpoint = None
         # list of GPU devices that are available and unused. If no GPUs, returns empty list
-        gpu_devices_dict = get_device_ids(num_devices)
+        gpu_devices_dict = get_device_ids(num_devices, max_used_perc=100, max_used_ram_perc=100)  # FIXME: set back to default after issue #246
```
remtav, line 247: the gpus don't empty between each run, so gpu usage threshold must be 100% otherwise each training after the first one runs on cpu (ie gpus get excluded)

mpelchat04: Or `get_device_ids` should be called once, before all runs.

mpelchat04 (dismissed):

victorlazio109 (dismissed):
Can we get a mini tutorial on how optuna works?

config/model/deeplabv3_pretrained.yaml
```diff
@@ -1,5 +1,5 @@
 # @package _global_
 model:
-  model_name: deeplabv3_resnet101
+  model_name: deeplabv3_pretrained
```
CharlesAuthier, line 3: I think that keeping the encoder in the name can be nice

CharlesAuthier (approved):

CharlesAuthier (commented):
When it will pass the test I will approve

remtav:
It seems geo-deep-learning will rely more and more on pytorch lightning (PL) and torchgeo (TG) for training. The hyperparameter optimization add-on is not a priority until the transition to PL and TG is cleared.

PR #286, opened 2022-03-07
fix a bug on calling  the dontcare value in dataset

CharlesAuthier:
Fixing a bug about the `dontcare` value from the segmentation dataset, this value is call in [train_segmentation](https://github.com/NRCan/geo-deep-learning/blob/develop/train_segmentation.py) line [411](https://github.com/NRCan/geo-deep-learning/blob/37cd99dc016df873ef363a1092e41e01b50eff42/train_segmentation.py#L411) and [416](https://github.com/NRCan/geo-deep-learning/blob/37cd99dc016df873ef363a1092e41e01b50eff42/train_segmentation.py#L416). The value was not set in the dataset I presume it was remove in an other PR and the code didn't follow, but this value is needed if we want to have the `iou` and  the `report_classification` save in mlflow or other trackers.

### DIFF
--- a/train_segmentation.py
+++ b/train_segmentation.py
@@ -90,6 +90,7 @@ def create_dataloader(samples_folder: Path,
 
     for subset in ["trn", "val", "tst"]:
         datasets.append(dataset_constr(samples_folder, subset, num_bands,
+                                       dontcare=dontcare_val,
                                        max_sample_count=num_samples[subset],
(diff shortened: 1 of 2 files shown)


remtav:
Since we'll probably entirely refactor the training script sooner than later to use Pytorch Lightning, is this PR still relevant @CharlesAuthier ?

PR #294, opened 2022-03-21
refactor model_choice.py using solution in issue #292 and more

### LINKED ISSUE #292: Model_choice.py: refactor model, loss and optimizer instantiation and loading
Model_choice.py has gotten messy and cluttered over the years and needs a bit of refactoring. This refactoring should be done before addressing #246 and #152. 

# Current state of things
- [set_hyperparamers()](https://github.com/NRCan/geo-deep-learning/blob/develop/models/model_choice.py#L108) function has very vague prupose of "set[ting] hyperparameters based on values provided in yaml config file";
- [net()](https://github.com/NRCan/geo-deep-learning/blob/develop/models/model_choice.py#L153) function is supposed to "Define the neural net", but in reality it's an all-in-one vague function that does the following:
1. Defines net architecture;
2. Reads a checkpoint to memory with [load_checkpoint()](https://github.com/NRCan/geo-deep-learning/blob/63796c8d9e848ac5bc23786386a09e2e5a96430a/models/model_choice.py#L72) (from a .pth.tar file as created by torch.save);
3. Returns if net() is called from inference or continues with the following if net() is called from train mode:
4. If more than one gpu is requested, determines [which gpus are available](https://github.com/NRCan/geo-deep-learning/blob/develop/models/model_choice.py#L206) based on user-inputted threshold for GPU's available RAM and usage %;
3.  [Sets model to DataParallel](https://github.com/NRCan/geo-deep-learning/blob/develop/models/model_choice.py#L215) if more than one gpu is requested and available;
5. Sets main device with [set_device()](https://github.com/NRCan/geo-deep-learning/blob/develop/models/model_choice.py#L224) function;
5. [Pushes model to main device](https://github.com/NRCan/geo-deep-learning/blob/develop/models/model_choice.py#L225);
6. Calls set_hyperparamters() (see above);
7. [Pushes loss to device](https://github.com/NRCan/geo-deep-learning/blob/develop/models/model_choice.py#L234);
8. Returns 7 (!!) objects: model, model_name, loss, etc.

# Suggested solution (high level)
All these steps could be better separated in small, dedicated functions of their own:
- read_checkpoint(): renamed version of load_checkpoint (prevents confusion with load_state_dict function). Although it derives from torch.load(checkpoint)'s function, this function really just reads a checkpoint in memory from a .pth.tar file to a Python dict containing weights, optimizer, etc.
- define_net_architecture(): define the model architecture from config parameters (i.e. create model with randomly initialized weights)
- adapt_checkpoint_to_dp_model(): for use at test loop during training only, adapts a generic checkpoint to be loaded to a DataParallel model as is done in [load_from_checkpoint](https://github.com/NRCan/geo-deep-learning/blob/develop/utils/utils.py#L65) (if model is DataParallel object)
- define_loss(): calls [verify_weights()](https://github.com/NRCan/geo-deep-learning/blob/develop/models/model_choice.py#L94) and instantiates a loss criterion
- define_optimizer(): instantiates optimizer with learning rate, weight decay, etc.

These functions would be called only when necessary in 3 main places:
1. Beginning of train_segmentation: 
- read checkpoint to be [loaded checkpoint for model weights and optimizer](https://pytorch.org/tutorials/beginner/saving_loading_models.html#load);
- define net architecture;
- load model weights with pytorch's [model_object].load_state_dict() method;
- define loss;
- define optimizer;
- load optimizer from checkpoint;
2. Test loop in train_segmentation: 
- Load best checkpoint to model (adapt checkpoint keys if model is a nn.DataParallel instance using dedicated function);
3. Beginning of inference: 
- override architecture, input bands, output classes from checkpoint's params;
- define net architecture
- load weights from provided checkpoint to model using pytorch's [model_object].load_state_dict() method

remtav:
- adapt train_segmentation.py and inference_segmentation.py to new usage
- move state_dict_path param to default_training.yaml
- implement unit tests for model_choice.py functions
- read_checkpoint(): add robustness (covers external checkpoints with only model weights, and complies to torch's save key naming standard 'model_state_dict' and 'optimizer_state_dict' rather than gdl's 'model' and 'optimizer' keys while being retrocompatible to old gdl models
- create high level define_model() function using all low level models definition/loading functions from model_choice.py
- test_losses.py: implement class weights test
- softcode strict loading boolean for loading provided state_dict at train_segmentation.py

fixes #292

### DIFF
--- a/config/gdl_config_template.yaml
+++ b/config/gdl_config_template.yaml
@@ -32,7 +32,6 @@ general:
   raw_data_dir: data
   raw_data_csv: tests/sampling/sampling_segmentation_binary_ci.csv
   sample_data_dir: data # where the hdf5 will be saved
-  state_dict_path:
   save_weights_dir: saved_model/${general.project_name}
 
 AWS:
--- a/config/model/deeplabv3_resnet101_dualhead.yaml
+++ b/config/model/deeplabv3_resnet101_dualhead.yaml
@@ -1,5 +1,6 @@
 # @package _global_
 model:
   model_name: deeplabv3_resnet101_dualhead
+  conc_point: conv1
   pretrained: True
   
--- a/config/training/default_training.yaml
+++ b/config/training/default_training.yaml
@@ -11,7 +11,8 @@ training:
   mode: 'min'            # 'min' or 'max', will minimize or maximize the chosen metric
   max_used_ram:
   max_used_perc:
-  target_size:
+  state_dict_path:
+  state_dict_strict_load: True
 
 #  precision: 16
 #  step_size: 4
--- a/inference_segmentation.py
+++ b/inference_segmentation.py
@@ -20,9 +20,9 @@ from pathlib import Path
 from omegaconf.listconfig import ListConfig
 
 from utils.logger import get_logger, set_tracker
-from models.model_choice import net
+from models.model_choice import define_model
 from utils import augmentation
-from utils.utils import load_from_checkpoint, get_device_ids, get_key_def, \
+from utils.utils import get_device_ids, get_key_def, \
     list_input_images, add_metadata_from_raster_to_sample, _window_2D, read_modalities, set_device
 from utils.verifications import validate_input_imagery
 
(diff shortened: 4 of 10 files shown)


CharlesAuthier (approved):

victorlazio109 (approved):

victorlazio109 (approved):
awesome cleanup!

mpelchat04 (approved):

PR #295, opened 2022-03-22
Use hydra for instantiating models and optimizers

remtav:
fixes #246 and #293 (breaking our one issue / one PR rule, forgive me!)
add tests for optimizer instantiation in test_optimizers.py
adapt our unet models (models/unet.py) to expect same parameter names as smp models

PR #297, opened 2022-03-23
Inference: override model and dataset parameters from parameters in checkpoint

### LINKED ISSUE #183: Inference: run inference without yaml
If #154 is covered, then we can easily adapt inference.py to run without a yaml. We could think of 2 mandatory arguments:
1. image url (local path or from web)
2. model url (local path or from web to checkpoint.pth.tar)

There could also be a third, optional argument:
- number of gpus. 

If this third argument is optional though, we'd have to determine what the default behavior is:
- if multiple gpus are available, run on one gpu only?
- if multiple gpus are available, run on all available gpus? This implies that #182 would be solved.

remtav:
[Draft until #295 is merged. This PR builds on top of #295.]

fixes #183 

This feature will make it possible once again to launch an inference from 2 simple arguments:
1. path/url to image
2. path/url to model checkpoint


remtav:
Will reopen to clean PR

PR #304, opened 2022-05-10
Update (remove unused columns, add aoi_id, reorganize columns) in input csv for tiling/sampling and inference

### LINKED ISSUE #221: Csv as input for sampling and inference: format needs update
STAC items are coming GDL's way and it's time to consider deprecating csv as input in favor of a STAC item, json-like, input. Jsons may not be as friendly if a user needs to create a custom list of images to input, but a simple utility could be used to convert a csv to the json expected as input.

remtav:
in input csv:
- "metadata" and "attribute_name" columns are removed (2nd and 4th column)
- ground truth column becomes 2nd column (not 3th)
- dataset split ("trn" or "tst") becomes 3rd column (not 5th)
- A last column (4th) is added for an "aoi id". This last information is not currently used, but could be to print a detailed report of tiling, training and evaluation, or for easier debugging.

Also in this PR:
- update test data for binary segmentation to smaller clipped tiles (under 100 Ko rather than 1Mb);
- remove data in data/ (use only data in tests/data/...
- default tiling to 32x32 chips for this smaller dataset;
- add ValueError if is train or val dataloader is empty

For more info on original suggested solution (slightly modified after discussion), see: https://github.com/NRCan/geo-deep-learning/issues/221#issuecomment-1121310523
fixes #221


CharlesAuthier (commented):
pass the test and its ok with me

CharlesAuthier (approved):

config/dataset/test_ci_segmentation_binary.yaml
```diff
@@ -2,7 +2,7 @@
 dataset:
   # dataset-wide
   name:
-  input_dim: 256
+  input_dim: 32
```
CharlesAuthier, line 5: 32?

CharlesAuthier (approved):

config/dataset/test_ci_segmentation_binary.yaml
remtav: The test data is much smaller than previous data (current is around 100x100 pixels, previous was around 1000x1000). To test the training pipeline the input dimension was reduced to 32. If you prefer, we could also maintain 256 (or 512, which is our standard for complete trainings), and simply overwrite this parameter during testing in the CI pipeline.

mpelchat04 (approved):

PR #316, opened 2022-06-17
Add automatic mixed precision (amp) for inference

mpelchat04:
While we wait for Pytorch Lightning to be implemented, I tested the torch automatic mixed precision AMP on QC10P001, for inference. Times are for raster prediction only.  
Original time for inference : 61m.59s    
Time for inference with AMP : 37m43s   
  
Number of Pixels that are different between both predictions : 20 (out of 403 605 000 pixels total).   
It also reduces the GPU memory needed (slightly).  

Configuration:   
I used Vic's NRG hydro model and tested it on Velum.

### DIFF
--- a/inference_segmentation.py
+++ b/inference_segmentation.py
@@ -211,7 +211,8 @@ def segmentation(param,
         for transformer in transforms:
             # augment inputs
             augmented_input = transformer.augment_image(inputs)
-            augmented_output = model(augmented_input)
+            with torch.cuda.amp.autocast():
+                augmented_output = model(augmented_input)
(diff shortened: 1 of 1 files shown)


CharlesAuthier:
That's a big PR next time make a smaller one :P

CharlesAuthier (approved):

remtav:
Cool! Fortunately, it's already in the production pipeline (https://github.com/remtav/geo-deep-learning/blob/222-stac-item-input/inference_segmentation.py#L176)

remtav (approved):

PR #341, opened 2022-09-08
Add band selection and ordering from multiband source rasters

### LINKED ISSUE #312: Multiband rasters as input: improve band selection
Based on PR #309, a band selection should be possible for multiband rasters. An easy implementation would be to expect indexes instead of band names for the `dataset.bands` parameter. Rasterio can use this parameter directly to choose relevant bands. The user would be reponsible of knowing the band corresponding to each index. We'd add this information in our documentation.

This means we'd remove the limited [`BGR_to_RGB()`](https://github.com/NRCan/geo-deep-learning/blob/develop/utils/utils.py#L294) function. Also the band selection and ordering would happen in the AOI object construction, not as a transform [as with `BGR_to_RGB()`](https://github.com/NRCan/geo-deep-learning/blob/develop/utils/augmentation.py#L363).

remtav:
fixes #312 
remove legacy BGR_to_RGB parameter and augmentation.
Replace by band selection and ordering via the AOI object (as already used for band selection with singleband rasters)


CharlesAuthier (approved):

valhassan (commented):
Band Strings vs Band Integers, it seems more explainable to go with strings coupled with the naming convention of our single bands?

remtav:
> Band Strings vs Band Integers, it seems more explainable to go with strings coupled with the naming convention of our single bands?

@valhassan when possible, it does seem preferable to use letters to identify bands. However, when dealing with multiband rasters, the knowledge of which bands constitute the raster and their order relies on the user, so I don't see any easy way to let a user select a subset or reorder bands in a multiband without having to support the reading of metadata and xml of specific providers

valhassan (approved):

PR #343, opened 2022-09-12
Tiling: replace hdf5 chips to .geotiff and .geojson chips

### LINKED ISSUE #215: Images_to_samples.py (hdf5s) --> data_to_tiles.py (tiles as tifs and geojsons)

remtav:
fixes #215 
Implements a new script for tiling using the solaris library and replacing the use of patches saved as numpy arrays in HDF5 files. The outputted patches are .tif and .geojson files directly. The .geojson label, once tiled, is then burned to .tif for use with the training script.

The 2 following diagrams are not up to date, but give a general idea of the tiling script's flow and what Tiler class might look light:
classes_data_to_tile drawio
data_to_tiles_flow drawio


tiling_segmentation.py
```diff
@@ -1,5 +1,6 @@
 from datetime import datetime
 import multiprocessing
+from numbers import Number
```
CharlesAuthier, line 3: Why import NUmber?

config/dataset/test_ci_segmentation_binary.yaml
```diff
@@ -9,7 +9,7 @@ dataset:
   download_data: False
 
   # imagery
-  bands: [R, G, B]
+  bands: [1,2,3]
```
CharlesAuthier, line 12: maybe add the meaning of 1,2,3  which is which ?

tests/test_tiling_segmentation.py
```diff
@@ -260,8 +260,8 @@ def test_tiling_inference(self):
         cfg = DictConfig(cfg)
         tiling(cfg)
         out_tiles = [
-            (Path("data/tiles/test_inference/tiles32_RGBbands/trn/23322E759967N_clipped_1m_1of2/images"), 99),
-            (Path("data/tiles/test_inference/tiles32_RGBbands/inference/23322E759967N_clipped_1m_2of2/images"), 176),
+            (Path("data/tiles/test_inference/tiles32_123bands/trn/23322E759967N_clipped_1m_1of2/images"), 99),
+            (Path("data/tiles/test_inference/tiles32_123bands/inference/23322E759967N_clipped_1m_2of2/images"), 176),
```
CharlesAuthier, line 264: A little bit confusing for a notation if I was new I would not know what mean `123bands`

tiling_segmentation.py
remtav: Used in type hints. This class is more permissive. Suggests that input should be either a float or int without have to explicitly allow these two types.

config/dataset/test_ci_segmentation_binary.yaml
remtav: It's mentionned in the doc [here](https://github.com/NRCan/geo-deep-learning/tree/develop/dataset#single-band-vs-multi-band-imagery) already. Is that enough?

tests/test_tiling_segmentation.py
remtav: Since GDL cannot know which spectral bands are present in multibands files, I see no other way around...

CharlesAuthier (approved):

CharlesAuthier (dismissed):

ms888ekb:
If at least one of the previously tiled samples is open in QGIS or any other app, the tiling script fails to generate a new set of tiles because it tries to move the previous files first (which are not accessible).
Do we agree to accept it or we want a user to have a chance to observe current tiles and generate new ones at the same time?

ms888ekb:
Did you notice that the training, testing and validation tiles are being generated always the same?
I'm wondering why this happens. I even changed (in case) 
`np.random.seed(123)`
but this didn't help.

train_segmentation.py
```diff
@@ -59,26 +61,38 @@ def create_dataloader(samples_folder: Path,
...
                             before training
     :return: trn_dataloader, val_dataloader, tst_dataloader
     """
     if not samples_folder.is_dir():
-        raise logging.critical(FileNotFoundError(f'\nCould not locate: {samples_folder}'))
-    if not len([f for f in samples_folder.glob('**/*.hdf5')]) >= 1:
-        raise logging.critical(FileNotFoundError(f"\nCouldn't locate .hdf5 files in {samples_folder}"))
-    num_samples, samples_weight = get_num_samples(samples_path=samples_folder, params=cfg)
+        raise FileNotFoundError(f'Could not locate: {samples_folder}')
+    experiment_name = samples_folder.parent.stem
```
ms888ekb, line 77: this code returns incorrect project name "patches" when expected is "template_project"
thus, the following function get_num_samples() can't properly identify *.csv files with tile paths.

it can be fixed like this:
`experiment_name = samples_folder.stem`

remtav:
> If at least one of the previously tiled samples is open in QGIS or any other app, the tiling script fails to generate a new set of tiles because it tries to move the previous files first (which are not accessible). Do we agree to accept it or we want a user to have a chance to observe current tiles and generate new ones at the same time?

@ms888ekb Important to consider indeed that some patches may be open in QGIS. However, the "moving of pre-existing patches" only happens [if debug=True](https://github.com/remtav/geo-deep-learning/blob/215-tiling-light/tiling_segmentation.py#L148), otherwise the tiling script will throw a "FileExistsError"

train_segmentation.py
remtav: I'll investigate. Thanks.

remtav:
> Did you notice that the training, testing and validation tiles are being generated always the same? I'm wondering why this happens. I even changed (in case) `np.random.seed(123)` but this didn't help.

I also noticed the result was always the same, but changing [the seed](https://github.com/remtav/geo-deep-learning/blob/215-tiling-light/tiling_segmentation.py#L25) as you mentionned changed the result. Can you test this again and get back to me if you still cannot change the result by changing the seed? @ms888ekb

tiling_segmentation.py
```diff
@@ -1,305 +1,423 @@
...
+        """
+        if not aoi.raster:  # in case of multiprocessing
+            aoi.raster_open()
+
+        raster_tiler = tile.raster_tile.RasterTiler(dest_dir=out_img_dir,
+                                                    src_tile_size=(self.dest_patch_size, self.dest_patch_size),
+                                                    alpha=False,
+                                                    verbose=True)
+        # TODO: fix bug with solaris: add possibility to serve metadata as input to tile()
+        aoi.raster.driver = "GTiff" if aoi.raster.driver == 'VRT' else aoi.raster.driver
```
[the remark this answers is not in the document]
remtav: I'll update the comment above this line with the following:
```
# workaround for successful writing of patches. 'VRT' driver cannot be assigned to written files. Error:
# rasterio.errors.RasterioIOError: Read or write failed. Writing through VRTSourcedRasterBand is not supported.
aoi.raster.driver = "GTiff" if aoi.raster.driver == 'VRT' else aoi.raster.driver
```

ms888ekb:
> > Did you notice that the training, testing and validation tiles are being generated always the same? I'm wondering why this happens. I even changed (in case) `np.random.seed(123)` but this didn't help.
> 
> I also noticed the result was always the same, but changing [the seed](https://github.com/remtav/geo-deep-learning/blob/215-tiling-light/tiling_segmentation.py#L25) as you mentionned changed the result. Can you test this again and get back to me if you still cannot change the result by changing the seed? @ms888ekb

@remtav I tested it again and the patches changed. Don't understand why they didn't last time but whatever, thanks.

remtav:
@ms888ekb seems you closed the PR. Did you want to "approve" PR rather? Closing will cancel the PR entirely and make no change to "develop" branch.

CharlesAuthier (approved):

ms888ekb:
> @ms888ekb seems you closed the PR. Did you want to "approve" PR rather? Closing will cancel the PR entirely and make no change to "develop" branch.

sorry @remtav, didn't want to close it.

tests/test_tiling_segmentation.py
```diff
@@ -0,0 +1,299 @@
...
+from pathlib import Path
+
+import numpy as np
+import pytest
+import rasterio
+from omegaconf import DictConfig
+from torchgeo.datasets.utils import extract_archive
+
+from dataset.aoi import AOI
+from tiling_segmentation import annot_percent, main as tiling, Tiler
```
mpelchat04, line 12: Review this import.

train_segmentation.py
```diff
@@ -160,28 +176,41 @@ def get_num_samples(samples_path, params):
...
-                    unique_labels = np.unique(label)
-                    weights.append(''.join([str(int(i)) for i in unique_labels]))
-                    samples_weight = compute_sample_weight('balanced', weights)
+            with open(dataset_filepath, 'r') as datafile:
+                num_samples[dataset] = len(datafile.readlines())
+
+        with open(dataset_filepath, 'r') as datafile:
+            datalist = datafile.readlines()
+            if dataset == 'trn':
+                # FIXME: user should decide whether or not this is used (very time consuming)
```
mpelchat04, line 202: Should be configurable and should still be there if needed.

train_segmentation.py
```diff
@@ -150,38 +173,66 @@ def calc_eval_batchsize(gpu_devices_dict: dict, batch_size: int, sample_size: in
...
+                else:
+                    dontcare = get_key_def("ignore_index", params['dataset'], default=-1)
+                    for x in tqdm(range(num_samples[dataset]), desc="Computing sample weights"):
+                        label_file = datalist[x].split(';')[1]
+                        with rasterio.open(label_file, 'r') as label_handle:
+                            label = label_handle.read()
+                        label = np.where(label == dontcare, 0, label)
+                        unique_labels = np.unique(label)
+                        weights.append(''.join([str(int(i)) for i in unique_labels]))
+                        samples_weight = compute_sample_weight('balanced', weights)
```
remtav, line 233: @valhassan are you ok if we softcode this? It creates lots of overhead at train time and for a binary task with patches already filtered to discard "background only" patches, I think we can skip the weighting component during training. It is set to False by default, but can be used by set training.compute_sampler_weights=True in config/training.

I quickly tested the weighting from .tif patches and it seems to be working fine.

valhassan: patched it in a separate branch similar to what you have. Works!
code snippet attached for reference
image

CharlesAuthier (approved):

ms888ekb (approved):

mpelchat04 (approved):

PR #345, opened 2022-09-21
Tests for metrics calculation + Torchmetrics for IoU

mpelchat04:
Adds tests for metrics calculation (classification report and IoU).  

IoU now relies on Torchmetrics for calculation.   

Classification report still uses scikit-learn because torchmetrics does not support ignore_index == -1.


tests/utils/test_metrics.py
```diff
@@ -0,0 +1,170 @@
+from utils.metrics import create_metrics_dict, report_classification, iou
+import torch
+import pytest
+
+# Test arrays: [bs=2, h=2, w,2]
```
CharlesAuthier, line 5: do you mean w=2?

tests/utils/test_metrics.py
```diff
@@ -0,0 +1,170 @@
...
+        """Evaluate report classification. 
+        Multiclass, without ignore_index in array."""
+        t = init_tensors()
+        metrics_dict = create_metrics_dict(3)
+        metrics_dict = report_classification(t['pred_multi'], 
+                                             t['lbl_multi'], 
+                                             batch_size=2, 
+                                             metrics_dict=metrics_dict,
+                                             ignore_index=-1)
+        assert "{:.6f}".format(metrics_dict['precision'].val) == "0.327083"
```
CharlesAuthier, line 50: Do we want a log of this in case of test on others machines like HPC?

CharlesAuthier (approved):

tests/utils/test_metrics.py
remtav: I don't see the necessity of logs for unit tests. Pytest does output a pretty detailed log if 
tests fail...

tests/utils/test_metrics.py
```diff
@@ -0,0 +1,170 @@
...
+        t = init_tensors()
+        metrics_dict = create_metrics_dict(3)
+        metrics_dict = report_classification(t['pred_multi'], 
+                                             t['lbl_multi_dc'], 
+                                             batch_size=2, 
+                                             metrics_dict=metrics_dict,
+                                             ignore_index=-1)
+        assert "{:.6f}".format(metrics_dict['precision'].val) == "0.297619"
+        assert "{:.6f}".format(metrics_dict['recall'].val) == "0.285714"
+        assert "{:.6f}".format(metrics_dict['fscore'].val) == "0.283163"
```
remtav, line 66: If you'd like to one step further in this verification, you could define some functions for precision, recall, fscore and iou from numpy directly and compare results from torchmetrics with these simple functions. This assertion is enough in my opinion though. The only downside is that it doesn't "prove" that a metric is being calculated the right way, it just proves that it's constant.

utils/metrics.py
```diff
@@ -1,17 +1,25 @@
...
+from math import sqrt
+from torch import IntTensor
+from torchmetrics import JaccardIndex
 
 min_val = 1e-6
-def create_metrics_dict(num_classes):
-    num_classes = num_classes if num_classes == 1 else num_classes + 1
+def create_metrics_dict(num_classes, ignore_index=None):
+
+    num_classes = num_classes + 1 if num_classes == 1 else num_classes
```
remtav, line 12: That's a pretty darn subtle bug :(
High five to the digital accelerator folks...

remtav (approved):

tests/utils/test_metrics.py
mpelchat04: Damn it! En effet.

tests/utils/test_metrics.py
mpelchat04: Right. I took the easy path of calculating the values in an excel sheet, to validate them.

utils/metrics.py
mpelchat04: Yup. But with the tests, we'll see these errors now!

CharlesAuthier (approved):

tests/utils/test_metrics.py
remtav: That's perfect. Thanks for validating!

remtav (approved):

PR #379, opened 2022-11-04
applie mask on inference

### LINKED ISSUE #346: Use NODATA mask to avoid out of bounds extractions
The extraction of our models is done on the whole extent of the images without considering the NODATA. Frequently we get false positives - especially from water -  in the out of bounds area since the band used for extraction exploits the entire 0-255 value range without dedicating any slot to the NODATA. It would be preferable to mask the extractions outside the valid area of the image.

image (3)

In the STAC storage structure that is used as input for the extraction, each "single-band COG" actually includes 2 bands - one for the actual 8-bit dataset and one for its mask.

The mask band is a 8-bit binary array with 0s and 255s
 
mask_image

**gdalinfo** 
> Band 1 Block=512x512 Type=Byte, ColorInterp=Gray
>   Overviews: 11214x16710, 5607x8355, 2803x4177, 1401x2088, 700x1044, 350x522, 175x261
>   Mask Flags: PER_DATASET
>   Overviews of mask band: 11214x16710, 5607x8355, 2803x4177, 1401x2088, 700x1044, 350x522, 175x261

**rasterio**
> import rasterio
> src = rasterio.open("/path/to/data")
> msk = src.read_masks(1)

My suggestion would be to append a step by the end of the inference process to apply the mask to the extraction raster to remove any out of bounds features.

CharlesAuthier:
Instead of clipping the prediction with a polygon mask this way we only force the value of the prediction to zero when outside the mask. After reading this doc (https://rasterio.readthedocs.io/en/latest/topics/masks.html) we can see that 0 value will be interpreted as no value. 

fixes #346

### DIFF
--- a/utils/geoutils.py
+++ b/utils/geoutils.py
@@ -225,3 +225,27 @@ def overlap_poly1_rto_poly2(polygon1: Polygon, polygon2: Polygon) -> float:
     """Calculate intersection of extents from polygon 1 and 2 over extent of a polygon 2"""
     intersection = polygon1.intersection(polygon2).area
     return intersection / (polygon2.area + 1e-30)
+
+
+def multi2poly(returned_vector_pred, layer_name):
+    """
+    Convert shapely multipolygon to polygon. If fail return a logging error.
+    This function will read an PATH string create an geodataframe and explode
+    all multipolygon to polygon and save the geodataframe at the same PATH.
+    Args:
+        returned_vector_pred: string, geopackage PATH where the post-processing
+                              results are saved
+        layer_name: string, the name of layer to look into for multipolygon, the name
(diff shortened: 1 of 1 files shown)


inference_segmentation.py
```diff
@@ -404,6 +404,10 @@ def main(params: Union[DictConfig, dict]) -> None:
                             tp_mem=temp_file,
                             debug=debug)
 
+        # applie the mask 
+        mask = aoi.mask
+        pred[mask == 0] = 0
```
ms888ekb, line 409: can we write:
`pred[aoi.mask == 0] = 0 `
here?

CharlesAuthier: oups hahahah thanks

PR #384, opened 2022-11-07
AOI: refactor "write destination multband raster from VRT to disk"

### LINKED ISSUE #380: [BUG] AOI: cannot write multiband destination raster in __init__()
Passing "virtual=False" will use a self.raster variable, but it is not yet assigned.
https://github.com/NRCan/geo-deep-learning/blob/develop/dataset/aoi.py#L209

remtav:
fixes #380 
This feature consists of letting user optionnally write to disk (as .tif) the VRT created after band selection and ordering from source input rasters. It was initially developped very quickly fro debugging and demoing without clear need for it. Now, @ms888ekb is working on using tiling with torchgeo, and I'm working on using torchgeo dataset in inference. This feature has become more important to consider as a possible way to use torchgeo datasets out-of-the-box.


dataset/aoi.py
```diff
@@ -470,24 +489,6 @@ def calc_raster_stats(self):
...
-            logging.error(f"To write a multi-band raster from single-band files, a VRT must be provided."
-                          f"\nGot {self.raster.meta}")
-            return
-        logging.debug(f"Writing multi-band raster to {self.raster_name}")
-        create_new_raster_from_base(
-            input_raster=self.raster,
-            output_raster=self.raster_name,
-            write_array=self.raster.read())
-        return self.raster_name
-
```
remtav, line 490: remove this function as "create_new_raster_from_base" can be used directly

tests/dataset/test_aoi.py
```diff
@@ -108,7 +108,10 @@ def test_stac_url_input(self):
         assert aoi.download_data is True
         assert Path("data/SpaceNet_AOI_2_Las_Vegas-056155973080_01_P001-WV03-R.tif").is_file()
         aoi.close_raster()
-        os.remove("data/SpaceNet_AOI_2_Las_Vegas-056155973080_01_P001-WV03-R.tif")
+        try:
+            os.remove("data/SpaceNet_AOI_2_Las_Vegas-056155973080_01_P001-WV03-R.tif")
+        except PermissionError:
+            pass
```
remtav, line 114: Fix for Windows

CharlesAuthier (dismissed):

tests/dataset/test_aoi.py
ms888ekb: Is there a test for the writing a multiband raster from VRT module?

config/tiling/README.md
```diff
@@ -53,3 +53,8 @@ particular patch, it is recommended to open the .geotiff version of ground truth
 If True, the tiling script uses Python's multiprocessing capability to process each AOI in parallel. This greatly 
 accelerates the tiling process. For testing or debugging purposes or for small dataset, we'd recommend keeping the 
 default `multiprocessing == False`.
+
+## Write destination multiband raster
+
+If True, will write the destination multiband raster from a VRT. Only affects cases when the source raster is composed 
+of singleband files being combined or if using a subset of bands from a multiband raster.
```
ms888ekb, line 60: Have two questions:
1. Current tests show that the destination multiband raster is written to the ./data folder anyways, even when the source raster is initially a multiband raster (not composed of several single-band rasters or not a subset of bands of a multiband raster) and I am personally OK with that. 
2. Do we want to write this dst multiband raster inside the associated project folder? currently it is in the ./data folder, as mentioned.

tests/dataset/test_aoi.py
remtav: Yes, with this test: https://github.com/NRCan/geo-deep-learning/blob/develop/tests/dataset/test_aoi.py#L258

config/tiling/README.md
remtav: 1. solved. If a source raster is fine as is, I don't see any need to create a VRT. Thanks for pointing that out. Implementation now follows what the documentation says.
2.  Currently, this folder is set in tiling_segmentation but the "raw_data_dir" parameter. This could be changed with the "experiment_name" directory. However, it seems more logical to me that an enhance copy of a source raster should belong to a directory where the "raw data" is, rather than a folder specific to the current experiment. There may not be a consensus about this in the team and I'm certainly open to discussing this. In any case, a user wanting the enhanced version to be saved within the experiment folder could simply point the "raw data dir" to the same folder as the "experiment name", rather than ./data.
Let me know if this all makes sense.

CharlesAuthier (dismissed):

dataset/aoi.py
```diff
@@ -157,7 +157,7 @@ def __init__(self, raster: Union[Path, str],
             If True, no rasterio.DatasetReader will be generated in __init__. User will have to call read raster later.
```
ms888ekb, line 157: looks the docstring paramenert need to be adjusted:
@param write_multiband -> @param write_dest_raster

tests/dataset/test_aoi.py
```diff
@@ -261,7 +264,7 @@ def test_write_multiband_from_single_band(self) -> None:
         data = read_csv("tests/tiling/tiling_segmentation_binary-singleband_ci.csv")
         row = data[0]
         aoi = AOI(raster=row['tif'], label=row['gpkg'], split=row['split'], raster_bands_request=['R', 'G', 'B'],
-                  write_multiband=True, root_dir="data")
+                  write_dest_raster=True, root_dir="data")
         assert Path("data/SpaceNet_AOI_2_Las_Vegas-056155973080_01_P001-WV03-R-G-B.tif").is_file()
         aoi.close_raster()
```
ms888ekb, line 269: I would add one more assertion here:
assert rasterio.open("data/SpaceNet_AOI_2_Las_Vegas-056155973080_01_P001-WV03-R-G-B.tif").count == 3

This way we also check:
1. The resulted raster is a valid GeoTIFF
2. All three bands were written.

ms888ekb (approved):

mpelchat04 (dismissed):

tiling_segmentation.py
```diff
@@ -121,14 +121,19 @@ def __init__(self,
                             f'Got {min_annot_perc} of type {type(min_annot_perc)}')
         self.min_annot_perc = min_annot_perc
 
-        bands_set = set([tuple(aoi.raster_bands_request) for aoi in self.src_aoi_list])
+        try:
+            bands_set = set([tuple(aoi.raster_bands_request) for aoi in self.src_aoi_list])
+            self.bands_requested = self.src_aoi_list[0].raster_bands_request
+        except TypeError:
+            bands_set = set([aoi.raster_meta['count'] for aoi in self.src_aoi_list])
+            self.bands_requested = list(range(1, self.src_aoi_list[0].raster_meta['count']+1))
```
remtav, line 129: Covers case if `raster_bands_request = [ ]`

CharlesAuthier (approved):

ms888ekb (approved):

ms888ekb (approved):

mpelchat04 (approved):

PR #388, opened 2022-11-14
Replace all references to hdf5 with "patches", remove hdf5 dependency and other minor refactoring

remtav:


config/README.md
```diff
@@ -84,8 +84,7 @@ general:
   min_epochs: 1 # for train only
   raw_data_dir: data
   raw_data_csv: tests/tiling/tiling_segmentation_binary_ci.csv
-  sample_data_dir: data # where the hdf5 will be saved
-  state_dict_path:
```
CharlesAuthier, line 88: Why removing `state_dict_path` ? This is not relate to hdf5, its for the pth file if I remember.

utils/create_dataset.py
```diff
@@ -31,39 +28,8 @@ def append_to_dataset(dataset, sample):
...
-        except AttributeError:
-            logging.exception(f'Update h5py to version 2.10 or higher')
-            raise
-        hdf5_files.append(hdf5_file)
-    return hdf5_files
-
-
 class SegmentationDataset(Dataset):
-    """Semantic segmentation dataset based on HDF5 parsing."""
+    """Semantic segmentation dataset based on input csvs listing pairs of imagery and ground truth patches as .tif."""
```
CharlesAuthier, line 32: csvs?

config/README.md
remtav: removing from README simply because this param no longers exists in our gdl_config_template: https://github.com/NRCan/geo-deep-learning/blob/develop/config/gdl_config_template.yaml#L35

utils/create_dataset.py
remtav: The SegmentationDataset expects as input a csv, not a path to hdf5 anymore: https://github.com/NRCan/geo-deep-learning/blob/develop/utils/create_dataset.py#L91
With current implementation, I'm noticing that the input csv could also be any text file formatted as a csv, but we could encourage the use of the csv by detailling the docs a bit more.

CharlesAuthier (approved):

mpelchat04 (approved):

Issue #404, opened 2022-12-02
Incorrect band subsetting for the multiband raster


ms888ekb:
I'm testing the visualization.py script, using this image:
Moncton2-011447328010_01_P002-WV02_red-green-blue_clahe10.tif
and corresponding label gpkg:
hydro_Moncton2-WV02-20180709-1.gpkg

It works now for RGB and 2 band images. However, when I tried to build a single-band dataset, i.e. subset only one band out of 3 available, it still generates 2-band tiles. The command I use:
python .\GDL.py mode=tiling dataset.attribute_field=class tiling.patch_size=512 dataset.attribute_values=[1] debug=False tiling.clahe_clip_limit=0 dataset.bands=[1]

UPDATE: The issue relates to the subset_multiband_vrt() function which does not work as expected.
The fixing patch is coming with the next PR.

PR #405, opened 2022-12-02
PR to fix #396 and #404

### LINKED ISSUE #396: Fix visualization for single-class mode
We would like to apply a threshold on outputs to get accurate visualization. 

Solution: apply threshold after line https://github.com/NRCan/geo-deep-learning/blob/8678cd900f12d2ded72395131b804c3babd74474/utils/visualization.py#L130

ms888ekb:
fixes #396, #404

Fixed vis() function
Fixed subset_multiband_vrt() function
Minor syntax fixes

### DIFF
--- a/tests/dataset/test_aoi.py
+++ b/tests/dataset/test_aoi.py
@@ -1,6 +1,7 @@
 import multiprocessing
 import os.path
 from pathlib import Path
+from typing import List
 
 import geopandas as gpd
 import numpy as np
@@ -38,12 +39,14 @@ class Test_AOI(object):
     @pytest.fixture(
         params=[[2, 1], [1], [2], [3], [1, 3]]
     )
-    def test_multiband_input_band_selection(self, request: SubRequest) -> None:
+    def bands_request(self, request: SubRequest) -> List:
+        return request.param
+
+    def test_multiband_input_band_selection(self, bands_request: List) -> None:
         """Tests reading a multiband raster as input with band selection"""
         extract_archive(src="tests/data/spacenet.zip")
         data = read_csv("tests/tiling/tiling_segmentation_binary-multiband_ci.csv")
         row = data[0]
-        bands_request = request.param
         aoi = AOI(raster=row['tif'], label=row['gpkg'], split=row['split'], raster_bands_request=bands_request)
         src_raster_subset = rasterio.open(aoi.raster_raw_input)
         src_np_subset = src_raster_subset.read(bands_request)
--- a/tests/dataset/test_aoi.py
+++ b/tests/dataset/test_aoi.py
@@ -5,6 +5,7 @@ from pathlib import Path
 import geopandas as gpd
 import numpy as np
 import pytest
+from _pytest.fixtures import SubRequest
 import rasterio
 from rasterio import RasterioIOError
 from shapely.geometry import box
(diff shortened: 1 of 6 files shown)


remtav (commented):
Thanks for addressing this.
Following our development workflow, I'd recommend splitting this PR into 2 separate PRs: 
1. bug fix for #396 
2. bug fix for subset_multiband_vrt()
For this second PR, first step would be to open an issue to document bug and provide minimal code to detect bug. For example, can you figure out why our [unit test](https://github.com/NRCan/geo-deep-learning/blob/develop/tests/dataset/test_aoi.py#L37) did not pick up on this bug? Can we modify this test to better cover this feature?

ms888ekb:
> 2\. bug fix for subset_multiband_vrt()
>     For this second PR, first step would be to open an issue to document bug and provide minimal code to detect bug. For example, can you figure out why our [unit test](https://github.com/NRCan/geo-deep-learning/blob/develop/tests/dataset/test_aoi.py#L37) did not pick up on this bug? Can we modify this test to better cover this feature?

This issue already exists:
https://github.com/NRCan/geo-deep-learning/issues/404

The test does not cover a case when we subset only one band (it covers only 2-band subsetting).
The testing unit is: test_multiband_input_band_selection() with the inputs:
bands_request = [2, 1]
If you change it to:
bands_request = [1]
it fails.

remtav (approved):

ms888ekb:
extended a unit test for the band selection function

tests/dataset/test_aoi.py
```diff
@@ -34,12 +35,15 @@ def test_multiband_input(self):
         aoi.close_raster()
         assert aoi.raster_closed
 
-    def test_multiband_input_band_selection(self):
+    @pytest.fixture(
+        params=[[2, 1], [1], [2], [3], [1, 3]]
+    )
+    def test_multiband_input_band_selection(self, request: SubRequest) -> None:
```
remtav, line 41: Cool!
@ms888ekb are you able to use Pycharm's debugger with the "test_multiband_input_band_selection" pytest fixture you added? I can't get it to work even though it works when debugging a very similarly structured test in torchgeo: https://github.com/microsoft/torchgeo/blob/main/tests/samplers/test_single.py#L94

Also, when I simply run tests from test_aoi.py, I get a total of 30 tests when the fixture is added, whereas I get total=31 without. It seems to be skipping this specific test when adding fixtures. I played around a bit to try and figure out why torchgeo's test works, but don't understand...

remtav: FYI, this is what I get when trying to run this specific test:
image

remtav: This works:
```
    @pytest.fixture(
        params=[[2, 1], [1], [2], [3], [1, 3]]
    )
    def bands_request(self, request: SubRequest) -> List:
        return request.param

    def test_multiband_input_band_selection(self, bands_request):
        """Tests reading a multiband raster as input with band selection"""
        extract_archive(src="tests/data/spacenet.zip")
        data = read_csv("tests/tiling/tiling_segmentation_binary-multiband_ci.csv")
        row = data[0]
        aoi = AOI(raster=row['tif'], label=row['gpkg'], split=row['split'], raster_bands_request=bands_request)
        ...
```

ms888ekb: Submitted the fix

remtav (approved):

CharlesAuthier (approved):

PR #425, opened 2022-12-08
Verify: use 2 metrics to check overlap between raster and label, not just iou

### LINKED ISSUE #410: AOI's bounds iou: better metric(s) to validate overlap between imagery and ground truth
In the process of examining input data, the verify_segmentation.py creates AOIs, then uses these instances' attributes to output a report. One relevant information in this report is the information concerning overlap between ground truth and imagery.

The [IoU metrics](https://github.com/NRCan/geo-deep-learning/blob/develop/dataset/aoi.py#L265) between ground truth and imagery cannot convey the relevant information about how ground truth overlaps imagery. We should separate this into 2 metrics:
1. % overlap of ground truth realtive to imagery's extent (ie intersection over imagery's extent) 
2. % of ground truth's extent that is inside imagery's extent. 

The first metric is interesting to better understand the data we're working with, but a % below 100% is not necessarily problematic (ex.: annotations may have covered the imagery only partially)

If this second metric's value is lower than 100%, then we have a serious issue with this particular AOI as some label data is located outside imagery. An error should be raised and user should investigate.

remtav:
fixes #410 

aoi.py: 
- set all attributes to None at beginning of \_\_init__
- store raster and label bounds as attribute
- calculate overlap of label relative to raster and vice-versa
- to_dict() outputs overlaps (for verify report)

geoutils.py: move bounds and overlap utils to geoutils.py
test_geoutils.py: move test from aoi to test_geoutils.py and update according to new metrics


CharlesAuthier (approved):

dataset/aoi.py
```diff
@@ -167,10 +167,15 @@ def __init__(self,
             https://scikit-image.org/docs/stable/api/skimage.exposure.html#skimage.exposure.equalize_adapthist
             https://kornia.readthedocs.io/en/latest/enhance.html#kornia.enhance.equalize_clahe
         """
-        self.raster_name_clahe = None
-        self.raster_dest = None
-        self.raster_np = None
+        self.raster = self.raster_dest = self.raster_np = self.raster_meta = self.raster_parsed = None
```
ms888ekb, line 170: this type of attributes initialization if confusing

ms888ekb (approved):

PR #426, opened 2022-12-08
add clipping gpkg to not go over the image(s)

CharlesAuthier:
1. Download the footprint from the stac-item during the inference, if included.
2. Clip the predicted gpkg outside the footprint after the postprocessing.

### DIFF
--- a/postprocess_segmentation.py
+++ b/postprocess_segmentation.py
@@ -6,6 +6,7 @@
 """CCMEO postprocess script."""
 import codecs
 import os
+import re
 import shutil
 import subprocess
 from typing import Dict, Union
@@ -505,8 +506,7 @@ def main(params):
             logging.critical(f'\nThe raw prediction contain no polygon.')
         else:
             gdf_footprint = geopandas.read_file(footprint)
-            gdf_footprint = gdf_footprint.to_crs(str(gdf_poly.crs))
-            
+            gdf_footprint = gdf_footprint.to_crs(str(gdf_poly.crs))            
             gdf_clipped = geopandas.clip(gdf_poly, gdf_footprint)
             gdf_clipped.to_file(returned_vector_pred, driver="GPKG")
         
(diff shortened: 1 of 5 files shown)


mpelchat04:
Could you include a test for this, along with the code changes?

remtav (commented):
Could we merge this in develop directly? After discussion with @mpelchat04 and seing the features @valhassan wants to bring into inference to make extraction better and faster, develop will most likely go into production soon, rathen a new version of the 222 branch. I can take care of merging develop into 222 until then. I think I'd make more sense to merge into develop though.

CharlesAuthier:
For now develop don't have a post-processing script, so not really.

CharlesAuthier:
Add tag item in the tiff image created from the inference and created a new layer inside the gpkg after the post-processing with all information wanted.

CharlesAuthier:
Add the function to convert multipolygon to polygon. @mpelchat04

mpelchat04 (approved):

remtav (changes requested):
Thanks for tackling these issues.
I understand this PR will also tackle adding tags to the final gpkg, right?
I see some discrepancies with our development workflow in this PR:
- Could you create an issue for this "tag to gpkg" feature?
- This second feature would normally be tackled by a branch/PR of its own. Is there a reason why this PR incorporates the bug fix and new feature at the same time?
- Could you link relevant issues to this PR?
- Could you add tests for the relevant areas of your code that should be tested? These tests will also help the reimplementation once we switch the produciton to the develop branch.

Let me know if I can help!

postprocess_segmentation.py
```diff
@@ -498,12 +524,77 @@ def main(params):
...
             returned_vector_pred = out_gen
+            
+            try: # Try to convert multipolygon to polygon
+                df = geopandas.read_file(returned_vector_pred, layer=layer_name)
+                if 'MultiPolygon' in df['geometry'].geom_type.values:
+                    logging.info("\nConverting multiPolygon to Polygon...")
+                    gdf_exploded = df.explode(index_parts=True, ignore_index=True)
+                    gdf_exploded.to_file(returned_vector_pred, layer=layer_name)
+            except Exception as e:
+                logging.error(f"\nSomething went wrong during the convertion of Polygon. \nError {type(e)}: {e}")
```
remtav, line 545: Could this become a function of its own?

postprocess_segmentation.py
```diff
@@ -498,12 +524,77 @@ def main(params):
...
+                    # Create a new layer in the gpkg with all information on stac item, gdl image and model class
+                    with fiona.open(returned_vector_pred, 'w', crs=_gpkg.crs, layer='extent_info', schema=extent_schema, driver='GPKG') as ds:
+                        p = {
+                            "stac_item": item_url,
+                            "gdl_image": reg_cont_image,
+                        }
+                        # add the checkpoint path associated with the model use for that class
+                        p.update({"model_"+cls_name : checkpoint_path})
+                        ds.write({'properties': p})
+
```
remtav, line 594: This is also a lengthy piece of code, could this become a function of its own? The dosctrings of this potential function would help quickly understand what's going on here and what are inputs/outputs.

postprocess_segmentation.py
CharlesAuthier: done

PR #437, opened 2022-12-15
inference_segmentation.py: add support for stac item as direct input

remtav:
Related to #222 

aoi.py: move stac class to its own file (dataset/stacitem.py)
test_stacitem.py: move stac-specific tests from test_aoi.py to test_stacitem.py

### DIFF
--- a/dataset/stacitem.py
+++ b/dataset/stacitem.py
@@ -1,0 +1,112 @@
+from collections import OrderedDict
+import logging
+from typing import Optional, Sequence, Dict, List
+
+import pystac
+from pystac.extensions.eo import ItemEOExtension, Band
+
+from utils.geoutils import is_stac_item
+
+logging.getLogger(__name__)
+
+
+class SingleBandItemEO(ItemEOExtension):
+    """
+    Single-Band Stac Item with assets by common name.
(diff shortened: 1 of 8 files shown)


CharlesAuthier (approved):

dataset/stacitem.py
```diff
@@ -0,0 +1,112 @@
...
+        elif isinstance(input_band, str) and len(input_band) == 1:
+            for cname, short_name in bands_ref:
+                if input_band == short_name:
+                    return cname
+        elif isinstance(input_band, str) and len(input_band) > 1:
+            for cname, short_name in bands_ref:
+                if input_band == cname:
+                    return input_band
+        else:
+            raise ValueError(f"Cannot convert given band to valid stac common name. Got: {input_band}")
```
remtav, line 112: This is hacky but necessary for models that may have been trained with bands as integers, but used at inference with stac item imagery (needs common names to specificy which band to use). Any better ideas?

mpelchat04: Make sense for now.  
Training with bands as integers should include a dict-like mapping to identify bands by their common names.

mpelchat04 (approved):

ms888ekb (approved):

Issue #442, opened 2023-01-04
Tag to gpkg


CharlesAuthier:
Add information to the TIFF after inference and transfer this information on the GPKG after the post-processing.

Information add: 
- TIFF: path to the weights and model details use for inference
- GPKG: path to the weights and model details use for inference, path to the STAC item (if used) and the image use for the post-processing for each layer

PR #448, opened 2023-01-10
New Feature - function fetching tag information to raster

CharlesAuthier:
New feature, function added in geoutils.py to fetch information save into tag inside a rasterio file. 
Also save checkpoint inside the inference raster at the end of `inference_segmentation.py`. This part is temporary, the tag information will be use via `create_new_raster_from_base` in the future.

This is necessary to tackle the #442 issue otherwise the GPKG will miss information.

### DIFF
--- a/utils/geoutils.py
+++ b/utils/geoutils.py
@@ -227,16 +227,16 @@ def overlap_poly1_rto_poly2(polygon1: Polygon, polygon2: Polygon) -> float:
     return intersection / (polygon2.area + 1e-30)
 
 
-def multi2poly(returned_vector_pred, layer_name):
+def multi2poly(returned_vector_pred, layer_name=None):
     """
     Convert shapely multipolygon to polygon. If fail return a logging error.
     This function will read an PATH string create an geodataframe and explode
     all multipolygon to polygon and save the geodataframe at the same PATH.
     Args:
         returned_vector_pred: string, geopackage PATH where the post-processing
-                              results are saved
+                              results are saved.
         layer_name: string, the name of layer to look into for multipolygon, the name
-                    represente the classes post-processed
+                    represente the classes post-processed. Default None.
     Return:
         none
     """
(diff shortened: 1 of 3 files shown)


utils/geoutils.py
```diff
@@ -247,5 +253,27 @@ def multi2poly(returned_vector_pred, layer_name):
...
+    with rasterio.open(raster_path) as tiff_src:
+        tags = tiff_src.tags()
+    # check if the tiff have the wanted tag save in 
+    if tag_wanted in tags.keys():
+        return tags[tag_wanted]
+    else:
+        raise logging.error(
+            f"\nThe tag {tag_wanted} was not found in the {tags.keys()},"
+            f" try again with one inside that list."
+        )
```
remtav, line 278: I seem to remember "raise logging.error()" doesn't work. Can you check?
You might have to go for:
```
logging.error(...)
raise "Exception"
```

utils/geoutils.py
```diff
@@ -227,16 +232,17 @@ def overlap_poly1_rto_poly2(polygon1: Polygon, polygon2: Polygon) -> float:
...
     This function will read an PATH string create an geodataframe and explode
     all multipolygon to polygon and save the geodataframe at the same PATH.
     Args:
         returned_vector_pred: string, geopackage PATH where the post-processing
-                              results are saved
-        layer_name: string, the name of layer to look into for multipolygon, the name
-                    represente the classes post-processed
+                              results are saved.
+        layer_name (optional): string, the name of layer to look into for multipolygon, the name
+                    represente the classes post-processed. Default None.
```
remtav, line 244: Suggested changes:
```
Converts shapely multipolygon to polygon. If fails, returns a logging error.
    This function will read a PATH string, create a geodataframe, explode
    all multipolygon to polygon and save the geodataframe at the same PATH.
    Args:
        returned_vector_pred: string, geopackage PATH where the post-processing
                              results are saved.
        layer_name (optional): string, the name of layer to look into for multipolygons. For example, if using during post-processing, the layer name could represent the class name if class are stored in separate layers. Default None.
```

remtav (approved):

ms888ekb (approved):

PR #456, opened 2023-01-19
Inference: optionnaly save a rescaled heatmap from logits

### LINKED ISSUE #217: inference.py: keep raw heatmap to set confidence level for extracted features after post-processing
Inference values for each before using argmax() function will be useful for technicians for sort extracted features by confidence levels. Our inference script should let use decide to output the raw heatmap (after softmax), not just the final one where argmax() has been applied

remtav:
fixes #217


ms888ekb (approved):

inference_segmentation.py
```diff
@@ -225,26 +227,22 @@ def segmentation(param,
     del fp
 
     fp = np.memmap(tp_mem, dtype='float16', mode='r', shape=(tf_len, h_padded, w_padded, num_classes))
-    pred_img = np.zeros((h_padded, w_padded), dtype=np.uint8)
+    pred_heatmap = np.zeros((h_padded, w_padded, num_classes), dtype=np.float16)
```
valhassan, line 230: extra dimension and data type changes will impact memory. Is this change necessary for creating the heatmaps or can it be handled in the heatmap function?

inference_segmentation.py
```diff
@@ -225,26 +227,22 @@ def segmentation(param,
...
                          leave=False, total=total_inf_windows, desc="Writing to array"):
         arr1 = (fp[:, row:row + chunk_size, col:col + chunk_size, :]).max(axis=0)
         if single_class_mode:
             arr1 = sigmoid(arr1)
-            arr1 = (arr1 > threshold)
-            arr1 = np.squeeze(arr1, axis=2).astype(np.uint8)
-        else:
-            arr1 = np.argmax(arr1, axis=-1).astype(np.uint8)
-        pred_img[row:row + chunk_size, col:col + chunk_size] = arr1
+        pred_heatmap[row:row + chunk_size, col:col + chunk_size, :] = arr1
```
valhassan, line 237: the name change pred_img --> pred_heatmap is not consistent with the output of the segmentation function.

inference_segmentation.py
remtav: I'll test the memory usage difference before and after this PR, but I don't see how we can keep the information from logits (ie to build heatmap) while still thresholding `arr1 = (arr1 > threshold)` with or "argmaxing" at this stage. Do you see any alternatives?

inference_segmentation.py
remtav: Can the "segmentation function" become a function that returns a heatmap rather than final prediction itself? I'd add this in docstrings, but how is pred_heatmap not consistent with the ouput of this function if we decide that it now returns a heatmap?

inference_segmentation.py
```diff
@@ -225,26 +227,22 @@ def segmentation(param,
...
 
     end_seg = time.time() - start_seg
     logging.info('Segmentation operation completed in {:.0f}m {:.0f}s'.format(end_seg // 60, end_seg % 60))
 
     if debug:
-        logging.debug(f'Bin count of final output: {np.unique(pred_img, return_counts=True)}')
+        logging.debug(f'Bin count of final output: {np.unique(pred_heatmap, return_counts=True)}')
```
remtav, line 243: This will need to be adjusted as np.unique will go crazy on the amount of values contained in the pred_heatmap since this array is float16, not uint8.

valhassan (approved):

inference_segmentation.py
valhassan: I don't see any alternatives but this change almost fully negates the purpose of the temporary file for memory savings

inference_segmentation.py
valhassan: Yes, it is fine. I understand now that thresholding has been moved to 'class_from_heatmap',

inference_segmentation.py
remtav: I'd like to discuss this with you before we merge this PR.

CharlesAuthier (approved):

inference_segmentation.py
```diff
@@ -390,6 +394,9 @@ def main(params: Union[DictConfig, dict]) -> None:
     device = set_device(gpu_devices_dict=gpu_devices_dict)
 
     clahe_clip_limit = get_key_def('clahe_clip_limit', params['tiling'], expected_type=Number, default=0)
+    heatmap_dtype = get_key_def('heatmap_dtype', params['inference'], default=np.uint16)
```
remtav, line 397: @valhassan dtype will default to uint16 to capture more details from original logits (originally as float32 or 16)

inference_segmentation.py
```diff
@@ -225,26 +229,26 @@ def segmentation(param,
...
-            arr1 = (arr1 > threshold)
-            arr1 = np.squeeze(arr1, axis=2).astype(np.uint8)
-        else:
-            arr1 = np.argmax(arr1, axis=-1).astype(np.uint8)
-        pred_img[row:row + chunk_size, col:col + chunk_size] = arr1
+
+        heatmap_max = np.iinfo(heatmap_dtype).max
+        arr1 = stretch_heatmap(heatmap_arr=arr1, out_max=heatmap_max)
+
+        pred_heatmap[row:row + chunk_size, col:col + chunk_size, :] = arr1.astype(heatmap_dtype)
```
remtav, line 243: @valhassan this should be more memory efficient. Does this make sense? I'm still performing memory tests, but this implementation should be better.

CharlesAuthier (approved):

CharlesAuthier (dismissed):

inference_segmentation.py
remtav: This is solved

inference_segmentation.py
remtav: This is ok since datatype will not be unsigned integer.

valhassan (approved):

CharlesAuthier (approved):

PR #457, opened 2023-01-19
inference_segmentation.py: softcode download directory for checkpoint if url

### LINKED ISSUE #455: Inference: softcode checkpoint file download directory if checkpoint if url

remtav:
fixes #455 

default_multiclass.yaml: add checkpoint_dir param
default_binary.yaml: idem
model_choice.py:
- remove load state dict path from define_model
- read_checkpoint: set default "update=False", not True
train_segmentation.py: adjust usage of define_model
test_models.py: idem


ms888ekb (dismissed):

models/model_choice.py
```diff
@@ -138,7 +136,4 @@ def define_model(
     )
     model = to_dp_model(model=model, devices=devices[1:]) if len(devices) > 1 else model
     model.to(main_device)
-    if state_dict_path:
```
valhassan, line 141: In my opinion, the logic here works well, why are we taking it out?

remtav: With the feature from this PR, loading weights from checkpoint in this function with its current usage would mean:
1. [Reading checkpoint in main(), and optionnaly downloading weights to checkpoint_dir if url](https://github.com/NRCan/geo-deep-learning/pull/457/files#diff-b85c0e821d4ba746c7ae4190dbc4ebafd445d4231a8a1f9796cd238fd79cc052R342)
2. If checkpoint was url, read_checkpoint() would need to return not only the checkpoint dictionary containing params and weights, but also the new path to downloaded local checkpoint (not url) for further use by define_model(). 
3. Define_model() would then take this updated path to read checkpoint a second time into memory, then perform the "model.load_state_dict" operation from those weights. 
Since, all of this would require reading two times rathen than once the provided checkpoint, it wouldn't be very optimal in my opinion (and would also require more lines of code). 

With this in mind, this PR would remove the "model loading" step in define_model() and use define_model() to reallty just "define model's architecture and push it to device"

remtav: I realize this change might need some discussion and should be tackled by separate PR, prior to this one. @valhassan please review: https://github.com/NRCan/geo-deep-learning/pull/461

inference_segmentation.py
```diff
@@ -424,8 +428,8 @@ def main(params: Union[DictConfig, dict]) -> None:
     # LOOP THROUGH LIST OF INPUT IMAGES
     for aoi in tqdm(list_aois, desc='Inferring from images', position=0, leave=True):
         Path.mkdir(working_folder / aoi.raster_name.parent.name, parents=True, exist_ok=True)
-        inference_image = working_folder / aoi.raster_name.parent.name / f"{aoi.raster_name.stem}_inference.tif"
-        temp_file = working_folder / aoi.raster_name.parent.name / f"{aoi.raster_name.stem}.dat"
+        inference_image = working_folder / f"{aoi.raster_name.stem}_inference.tif"
```
remtav, line 431: waiting for #456 to add heatmap renaming here.

inference_segmentation.py
```diff
@@ -451,7 +455,7 @@ def main(params: Union[DictConfig, dict]) -> None:
             dest.write(pred)
             # NOTE: the tags option will be join to the `create_new_raster_from_base` function for later version
             # add tag to transmit more informations, the checkpoint path in that case
-            dest.update_tags(checkpoint=state_dict)  
+            dest.update_tags(checkpoint=state_dict)
```
remtav, line 458: This will also disappear with #456

valhassan (dismissed):

inference_segmentation.py
```diff
@@ -354,10 +360,6 @@ def main(params: Union[DictConfig, dict]) -> None:
     num_classes = num_classes + 1 if num_classes > 1 else num_classes  # multiclass account for background
     num_bands = len(bands_requested)
 
-    working_folder = state_dict.parent.joinpath(f'inference_{num_bands}bands')
-    logging.info("\nThe state dict path directory used '{}'".format(working_folder))
-    Path.mkdir(working_folder, parents=True, exist_ok=True)
-    logging.info(f'\nInferences will be saved to: {working_folder}\n\n')
```
remtav, line 360: The output folder will change to a more simple hierarchy. The destination folder will become the user's choice via the "inferences.root_dir" parameter.

valhassan: Can you confirm if evaluate_segmentation works with the new hierarchy?

remtav: Good point. I will check.

mpelchat04 (approved):

CharlesAuthier (approved):

PR #461, opened 2023-01-23
model_choice.py: remove define_model() function

### LINKED ISSUE #460: Model_choice.py: define_model() function is superfluous and adds constraints and overhead when loading pretrained weights
See [define_model()](https://github.com/NRCan/geo-deep-learning/blob/develop/models/model_choice.py#L121).

It doesn't seem relevant to add define_model() function to the already existing define_model_architecture(). What define_model() does is 
(1) convert model to a DataParallel model is more than 1 gpu is used (which is only relevant for training and can be performed directly in main() function)
(2) push model to device, which can (and should?) be done directly in main() function for training and inference (without adding lines of code really).
(3) read weights from a checkpoint file if this file is provided. However, this operation creates overhead at inference, since the checkpoint must already be read once in main() function to override default params.

Related to #455 ::
With the "softcode download directory for checkpoints that are urls" feature, loading weights from checkpoint in this function with its current usage would mean:
1. Reading checkpoint in main(), and optionnaly downloading weights to checkpoint_dir if url
2. If checkpoint was url, read_checkpoint() would need to return not only the checkpoint dictionary containing params and weights, but also the new path to downloaded local checkpoint (not url) for further use by define_model().
3. Define_model() would then take this updated path to read checkpoint a second time into memory, then perform the "model.load_state_dict" operation from those weights.

Since, all of this would require reading two times rathen than once the provided checkpoint, it wouldn't be very optimal in my opinion (and would also require more lines of code).

remtav:
Fixes #460

### DIFF
--- a/config/inference/default_binary.yaml
+++ b/config/inference/default_binary.yaml
@@ -1,8 +1,10 @@
 # @package _global_
 inference:
   raw_data_csv: tests/inference/inference_segmentation_binary.csv
+  root_dir: data/inference
   input_stac_item:  # alternatively, use a path or url to stac item directly
   state_dict_path: ${general.save_weights_dir}/
+  checkpoint_dir: ${general.save_weights_dir}  # (string, optional): directory in which to save the object if url
   chunk_size:  # if empty, will be calculated automatically from max_pix_per_mb_gpu
   # Maximum number of pixels each Mb of GPU Ram to allow. E.g. if GPU has 1000 Mb of Ram and this parameter is set to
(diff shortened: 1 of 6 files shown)


CharlesAuthier (approved):

ms888ekb (approved):

remtav:
@valhassan mentionned that define_model() seems still useful in train_segmentation.py to handle model architecture definition and weight loading into a single function. Define_model() will live, for now :)

PR #462, opened 2023-01-23
AOI: download asset only if not already downloading

### LINKED ISSUE #363: Competition problem: parallel inferences in production
In the production inference process, there is a step where we check the presence or not of the asset in the data repository to avoid downloading it in HPC every time.

In the context where we use a new image that has never been loaded in HPC and we want to run all our models in parallel, this has the side effect that the first of the N models downloads the assets in the data folder as expected, but other runs detect the presence of the TIFs while they are being written in said directory. This leads to read errors in the other steps of the process while some tiles are incomplete, etc.

We should find a more complete way to validate the presence of assets in HPC so that we don't include those in the process of downloading.

**Proposed implementation: loop over the file size.**

```
if not os.path.exists(val):
	# Download file
	
else:
    file_size = -1
	while file_size != os.path.getsize(val):
	    file_size = os.path.getsize(val)
		time.sleep(1)

     # Path exists and file size is stable. Good to go.
```

remtav:
fixes #363 
EDIT by Remi on 16/02/23:
I'm restoring this PR to simplify the conflicts when multiple inferences on a same image are ran at once.
This "bugfix" will exist in parallel with the previously added "inference.prep_data_only" (supposed to solve the same issue), since this last feature is very small and could be useful in the future.


ms888ekb (approved):

CharlesAuthier (dismissed):

CharlesAuthier (dismissed):

CharlesAuthier (dismissed):

remtav:
replaced with #471

CharlesAuthier (dismissed):

CharlesAuthier (dismissed):

utils/utils.py
```diff
@@ -643,6 +646,81 @@ def update_gdl_checkpoint(checkpoint: Union[dict, DictConfig]) -> Dict:
...
+        file.seek(0, 2)
+        # retrieve the current position of the pointer
+        # this will be the file's size in bytes
+        size = file.tell()
+        return size
+    # if the function reaches this statement it means an error occurred within the above context handler
+    return False
+
+
+def wait_while_modif(fpath: Union[str, Path], sleep_secs: int = 10, timeout: int = 1800) -> None:
```
ms888ekb, line 665: did you have unit tests for these new functions?

utils/utils.py
```diff
@@ -643,6 +646,81 @@ def update_gdl_checkpoint(checkpoint: Union[dict, DictConfig]) -> Dict:
...
+    Copied from https://instructobit.com/tutorial/117/Detecting-files-in-the-process-of-being-copied-or-written-to-in-Python#Detecting-changed-made-to-large-files-using-size-comparison
+    """
+    # open the file in read only
+    with open(filepath, "r") as file:
+        # move pointer to the end of the file
+        file.seek(0, 2)
+        # retrieve the current position of the pointer
+        # this will be the file's size in bytes
+        size = file.tell()
+        return size
```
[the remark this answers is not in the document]
remtav: No good reason. I just relied on an implementation found on internet. I guess it works too. I just changed it :) Less code is necessary this way.

utils/utils.py
remtav: Easy answer: yes and no.
Yes: I added a test `test_download_url_wcheck(self):` under test_utils.py that call this function.
No: no test for this function directly. I have a hard time figuring out how to test this. I'd have to find a way to modify a file in background while checking if it's being modified with this function. I find the test mentionned above sufficient...

CharlesAuthier (approved):

valhassan (approved):

mpelchat04 (approved):

PR #467, opened 2023-01-27
Fix to handle uint16 data and 'geometrycollection' issue

ms888ekb:


dataset/create_dataset.py
```diff
@@ -179,6 +179,11 @@ def _get_tensor(self, query):
             out_shape=out_shape, window=from_bounds(*bounds, self.dr_ds.transform)
         )
 
+        if dest.dtype == np.uint16:
+            dest = dest.astype(np.int32)
+        elif dest.dtype == np.uint32:
+            dest = dest.astype(np.int64)
+
```
CharlesAuthier, line 186: can you add a else with a log error pls

ms888ekb: Here can't be "else" because all other cases are handled except these two.
I borrowed this workaround from the official implementation of the TorchGeo's RasterDataset class.

valhassan (approved):

remtav (approved):

PR #478, opened 2023-02-14
Resolving the raising error when `get_key_def` not match the expected type

### LINKED ISSUE #476: expected_type argument in get_key_def() doesn't do anything
the expected_type in get_key_def() function may not have the intended behaviour (assume it's forcing the return value to have the expected_type), 

for example: learning_rate = get_key_def(['training', 'lr'], cfg, default=None, expected_type=str) would return a float and not raise any exception.

CharlesAuthier:
Fix #476 by moving the exception at the end of the function and adding a test to the mix.

### DIFF
--- a/inference_segmentation.py
+++ b/inference_segmentation.py
@@ -361,7 +361,7 @@ def main(params: Union[DictConfig, dict]) -> None:
     num_bands = len(bands_requested)
 
     # Default input directory based on default output directory
-    raw_data_csv = get_key_def('raw_data_csv', params['inference'], expected_type=str, to_path=True,
+    raw_data_csv = get_key_def('raw_data_csv', params['inference'], expected_type=Path, to_path=True,
                                validate_path_exists=True)
     input_stac_item = get_key_def('input_stac_item', params['inference'], expected_type=str, to_path=True,
                                   validate_path_exists=True)
--- a/tests/utils/test_utils.py
+++ b/tests/utils/test_utils.py
@@ -2,13 +2,14 @@ import pytest
 from torchgeo.datasets.utils import extract_archive
 
 from hydra import initialize, compose
-from omegaconf import DictConfig, OmegaConf, open_dict
+from omegaconf import OmegaConf
+import unittest
 
 from models.model_choice import read_checkpoint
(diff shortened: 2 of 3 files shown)


inference_segmentation.py
```diff
@@ -361,7 +361,7 @@ def main(params: Union[DictConfig, dict]) -> None:
     num_bands = len(bands_requested)
 
     # Default input directory based on default output directory
-    raw_data_csv = get_key_def('raw_data_csv', params['inference'], expected_type=str, to_path=True,
+    raw_data_csv = get_key_def('raw_data_csv', params['inference'], expected_type=Path, to_path=True,
```
ms888ekb, line 364: can "str "or "Path" types be both valid (expected) data types for the given parameter?

CharlesAuthier: Yeah but the parameter `to_path` is set to True so at the end you are suppose to have a Path

remtav: I follow @ms888ekb's thought: it seems a bit confusing that "raw_data_csv" can be a str, but the expected type is "Path". I understand when I look at the code modifications that even if a str is inputted, it will become a Path, but I feel like it could be confusing. Will still approve this PR since this is minor.

tests/utils/test_utils.py
```diff
@@ -46,3 +50,17 @@ def test_update_gdl_checkpoint(self) -> None:
...
+        with initialize(config_path="../../config", job_name="test_key_def"):
+            cfg = compose(config_name="gdl_config_template", return_hydra_config=True)
+            hconf = HydraConfig()
+            hconf.set_config(cfg)
+            # Not the same type - raise the error
+            with self.assertRaises(TypeError):
+                get_key_def('max_pix_per_mb_gpu', cfg['inference'], default=25, expected_type=str)
+            # Same type
+            mp = get_key_def('max_pix_per_mb_gpu', cfg['inference'], default=25, expected_type=int)
+            assert isinstance(mp, int)
```
remtav, line 64: Cool, thanks for adding this!

remtav (approved):

PR #480, opened 2023-02-15
nodata labels masking

ms888ekb:
- filters 100% nodata raster tiles and corresponding label
- reads nodata mask from the raster, convert it to vector polygon and clip corresponding gpkg 
- masks burned labels with ignor_index where raster is nodata.


CharlesAuthier (dismissed):

CharlesAuthier (approved):

CharlesAuthier (approved):

tests/dataset/test_datasets.py
```diff
@@ -50,6 +69,45 @@ def test_getitem(self, vector_dataset: GDLVectorDataset) -> None:
         assert isinstance(x["mask"], ogr.DataSource)
         assert isinstance(x['bbox'], BoundingBox)
 
+    def test_clip_nodata(self):
```
remtav, line 72: Nice addition, thanks

dataset/create_dataset.py
```diff
@@ -191,23 +191,34 @@ def _get_tensor(self, query):
...
         """
         super().__init__()
 
-        self.vec_ds = ogr.Open(str(vec_ds))
+        # Read the dataset as an OGR datasouce:
+        vec_ds = ogr.Open(str(vec_ds))
         self.res = res
 
-        assert self.vec_ds is not None, "The vector dataset is empty."
+        assert vec_ds is not None, "The vector dataset is empty."
```
remtav, line 210: ideally, a ValueError could be raised.

utils/geoutils.py
```diff
@@ -305,3 +306,88 @@ def gdf_mean_vertices_nb(gdf: gpd.GeoDataFrame):
...
+    Args:
+        img_patch: raster tile image path
+        gt_patch: raster tile label path
+        nodata_val: nodata value
+        mask_val: masking value (255 by default)
+
+    Returns:
+        Masks label tile or None if no nadata pixels
+    """
+    image_ds = gdal.Open(str(img_patch), gdalconst.GA_ReadOnly)
```
remtav, line 323: Ideally, GDAL could be repalced where possible with rasterio, as this is one of rasterio's main goal:
https://rasterio.readthedocs.io/en/latest/contributing.html#design-principles

According to rasterio's doc, there's also some risk in using both gdal and rasterio in our scripts: https://rasterio.readthedocs.io/en/latest/topics/switch.html#mutual-incompatibilities

ms888ekb: Particularly for this function, I was choosing between rasterio and gdal and found gdal implementation more concise and clear.
To mask the label patch in case of rasterio, you need to read the array, save a profile, then create a new file (which I don't want to do) and then save the data in it.
GDAL provides updatinf mechanism such as gdalconst.GA_Update, so you can just rewrite the array into the same file.

remtav (approved):
Very complete PR with nice tests and flexible input for nodata (mask and value).
Great work @ms888ekb .
Comments are mostly to leave some traces at this point, especially for the risk when using gdal and rasterio simultaneously.

tests/test_tiling_segmentation.py
```diff
@@ -37,7 +37,7 @@ def test_outputted_chips(self):
         tiling(cfg)
         # expected number of patches is constant due to random seed set in tiling script
         out_labels = [
-            (Path(f"{data_dir}/{proj}/trn/22978945_15_uint8_clipped/labels_burned"), 16),
+            (Path(f"{data_dir}/{proj}/trn/22978945_15_uint8_clipped/labels_burned"), 13),
```
remtav, line 40: How come the output amount of labels is less than before if imagery has no nodata?

mpelchat04 (approved):

PR #482, opened 2023-02-15
Inference: implement user-chosen output path

### LINKED ISSUE #472: Inference: softcode output name of extraction
Being able to define the output name of inference output from the configuration adds flexibility to the inference process. For example, for a given image, we may run multiple models and want to output name to bear the class identifier as suffix.

remtav:
fixes #472 (oups the PR branch has wrong number!)
inference_segmentation.py:
- implement output_path and name heatmap and temp file accordingly
- replace aoi.raster_name to aoi.aoi_id for cleaner log README.md: update inference and aoi READMEs
default_binary.yaml: add output_path
default_multiclass.yaml: idem

### DIFF
--- a/inference_segmentation.py
+++ b/inference_segmentation.py
@@ -469,7 +469,7 @@ def main(params: Union[DictConfig, dict]) -> None:
                          "count": pred_heatmap.shape[0],
                          "dtype": 'uint8',
                          "compress": 'lzw'})
-        logging.info(f'\nSuccessfully inferred on {aoi.aoi_id}\nWriting to file: {inference_image}')
+        logging.info(f'\nSuccessfully inferred on {aoi.aoi_id}\nWriting to file: {output_path}')
 
         pred_img = class_from_heatmap(heatmap_arr=pred_heatmap, heatmap_threshold=heatmap_threshold)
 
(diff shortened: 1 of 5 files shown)


inference_segmentation.py
```diff
@@ -437,10 +436,16 @@ def main(params: Union[DictConfig, dict]) -> None:
 
     # LOOP THROUGH LIST OF INPUT IMAGES
     for aoi in tqdm(list_aois, desc='Inferring from images', position=0, leave=True):
-        inference_image = working_folder / f"{aoi.raster_name.stem}_inference.tif"
-        inference_heatmap = working_folder / f"{aoi.raster_name.stem}_heatmap.tif"
-        temp_file = working_folder / f"{aoi.raster_name.stem}.dat"
-        logging.info(f'\nReading image: {aoi.raster_name.stem}')
+        inference_image = get_key_def(
```
CharlesAuthier, line 439: I dont know if looping on a `get_key_def` is the best solution, time wise I think it will be wiser to make the call outside the function and if `{aoi.aoi_id}_pred.tif` not there then reassign inference_image. 

But I can be wrong, its just that time wise calling a function is more time consuming that just checking if it there.

CharlesAuthier (dismissed):

CharlesAuthier (approved):

valhassan (approved):

PR #484, opened 2023-02-21
utils.py/get_key_def(): allow "expected_type=str" even with "to_path=True" param

remtav:
inference_segmentation.py: restore str as expected type to better reflect documentation

Documentation for "expected_type" in get_key_def(): [:param expected_type: (type) type of the expected variable.](https://github.com/NRCan/geo-deep-learning/blob/develop/utils/utils.py#L142)

To me, this suggests that the input variable must bear this "expected_type". It feels confusing to expected a type that is actually one that is assigned during the get_key_def() function, "Path" in this case.
@CharlesAuthier is this slight retroactive change ok to you?

### DIFF
--- a/utils/utils.py
+++ b/utils/utils.py
@@ -175,6 +175,7 @@ def get_key_def(key, config, default=None, expected_type=None, to_path: bool = F
             val = Path(to_absolute_path(val))
         except TypeError:
             logging.error(f"Couldn't convert value {val} to a pathlib.Path object")
+        expected_type = Path if expected_type == "str" else expected_type  # allows "str" and "Path" as expected_type
     if validate_path_exists:
         if not isinstance(val, Path):
             val = Path(to_absolute_path(val))
--- a/utils/utils.py
+++ b/utils/utils.py
@@ -175,7 +175,7 @@ def get_key_def(key, config, default=None, expected_type=None, to_path: bool = F
             val = Path(to_absolute_path(val))
         except TypeError:
             logging.error(f"Couldn't convert value {val} to a pathlib.Path object")
-        expected_type = Path if expected_type == "str" else expected_type  # allows "str" and "Path" as expected_type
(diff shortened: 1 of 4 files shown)


CharlesAuthier:
I dont care but change the test too if you want to pass the check

CharlesAuthier:
https://github.com/NRCan/geo-deep-learning/blob/develop/tests/utils/test_utils.py#L54

utils/utils.py
```diff
@@ -175,6 +175,7 @@ def get_key_def(key, config, default=None, expected_type=None, to_path: bool = F
             val = Path(to_absolute_path(val))
         except TypeError:
             logging.error(f"Couldn't convert value {val} to a pathlib.Path object")
+        expected_type = Path if expected_type == str else expected_type  # allows "str" and "Path" as expected_type
```
CharlesAuthier, line 178: change your `expected_type == str` for ` isinstance(expected_type, str)` it's possible that python dont like when you use `==` to compare type

CharlesAuthier (changes requested):

CharlesAuthier:
btw can you add a test to check that pls

utils/utils.py
remtav: Makes sense, but doesn't seem to be working for some reason:
image

ms888ekb (approved):

ms888ekb (approved):

valhassan (approved):

utils/utils.py
remtav: Actually, I understood: isinstance() check the type of str. It's type is a Python "type" object, so the "expected_type" is itself a type class of type "str", not a type class. That's super hard to explain and would need a bit more details, but not too important at this point. :)

CharlesAuthier (approved):

Issue #485, opened 2023-02-22
Smoothening with Hann Window


valhassan:
New Feature Request: We would like to smooth predictions at inference which solves edge effects, jaggedness, and holes when stitching tiles/patches. Below is an illustration of problem areas highlighted with red circles. 
1. Original Raster, 2. Water Body Extraction (No Smoothening), 3. Water Body Extraction (Smoothening with Hann).

![1](https://user-images.githubusercontent.com/24499815/224417512-27c080af-9918-4c06-9f62-3aaab8fd8945.png "Raster")
![tempsnip_2](https://user-images.githubusercontent.com/24499815/224417571-c5805eca-3640-4d3e-9d12-1ae74c1a0e34.png "No Smoothening")
![tempsnip_3](https://user-images.githubusercontent.com/24499815/224417597-8359cb28-b07a-4aa7-9b9d-6cb9af08ecda.png "Smoothening with Hann Window")

**Method**

1. Generate overlapping patches, with a stride half of the patch size.
2. Model prediction on patches.
3. Multiplying each patch by the appropriate window depending on its absolute location.
4. Stitching all patches at their absolute location.

Reference Paper: https://arxiv.org/abs/1910.07831

PR #486, opened 2023-02-22
Smoothening with Hann Window

valhassan:
minor fix

### DIFF
--- a/inference_segmentation.py
+++ b/inference_segmentation.py
@@ -13,6 +13,7 @@ import time
 import fiona  # keep this import. it sets GDAL_DATA to right value
 import rasterio
 import ttach as tta
+from scipy.special import softmax
 from collections import OrderedDict
 from fiona.crs import to_string
 from ruamel_yaml.comments import CommentedSeq
@@ -31,25 +32,23 @@ from utils.inference import stretch_heatmap, class_from_heatmap
 from utils.logger import get_logger, set_tracker
 from models.model_choice import define_model, read_checkpoint
 from utils import augmentation
+from utils.inference import generate_patch_list
 from utils.utils import get_device_ids, get_key_def, \
-    add_metadata_from_raster_to_sample, _window_2D, set_device
+add_metadata_from_raster_to_sample, set_device
 
 # Set the logging file
 logging = get_logger(__name__)
+
 def _pad(arr, chunk_size):
     """ Pads img_arr """
     w_diff = chunk_size - arr.shape[0]
     h_diff = chunk_size - arr.shape[1]
-    aug = int(round(chunk_size * (1 - 1.0 / 2.0)))
-    aug_w = int(round(chunk_size * (1 - 1.0 / 2.0))) + w_diff
-    aug_h = int(round(chunk_size * (1 - 1.0 / 2.0))) + h_diff
     if len(arr.shape) > 2:
-        padded_arr = np.pad(arr, ((aug, aug_w), (aug, aug_h), (0, 0)), mode="reflect")
+        padded_arr = np.pad(arr, ((0, w_diff), (0, h_diff), (0, 0)), mode="reflect")
     else:
-        padded_arr = np.pad(arr, ((aug, aug_w), (aug, aug_h)), mode="reflect")
+        padded_arr = np.pad(arr, ((0, w_diff), (0, h_diff)), mode="reflect")
 
     return padded_arr
-
 
 def ras2vec(raster_file, output_path):
     # Create a generic polygon schema for the output vector file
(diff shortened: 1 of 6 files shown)


CharlesAuthier (dismissed):

valhassan:
Please review the relevant issue here: https://github.com/NRCan/geo-deep-learning/issues/485

CharlesAuthier (approved):

inference_segmentation.py
```diff
@@ -157,32 +156,24 @@ def segmentation(param,
     Returns:
 
     """
-    subdiv = 2
+    use_hanning = True
```
remtav, line 159: Could this be softcoded? Even if we intend to always use hanning, it seems like it would be nice to set this value from config (default=True) at least to hint the user than some smoothening is going on.

inference_segmentation.py
```diff
@@ -157,32 +156,24 @@ def segmentation(param,
...
     print_log = True if logging.level == 20 else False  # 20 is INFO
-    pad = chunk_size * 2
-    h_padded, w_padded = [side + pad for side in input_image.shape]
-    dist_samples = int(round(chunk_size * (1 - 1.0 / 2.0)))
-
     model.eval()  # switch to evaluate mode
 
     # initialize test time augmentation
-    transforms = tta.Compose([tta.HorizontalFlip(), ])
+    transforms = tta.aliases.d4_transform()
```
remtav, line 166: Could you include this change in the PR description as this seems like it could have a big impact and it's not directly related to Hann window implementation

inference_segmentation.py
```diff
@@ -214,30 +205,25 @@ def segmentation(param,
...
-                         leave=False, total=total_inf_windows, desc="Writing to array"):
-        arr1 = (fp[:, row:row + chunk_size, col:col + chunk_size, :]).max(axis=0)
+    for row, col in tqdm(itertools.product(range(0, h_padded, chunk_size),
+                                           range(0, w_padded, chunk_size)),
+                         leave=False, position=0, desc="Writing to array"):
+        arr1 = (fp[:, row:row + chunk_size, col:col + chunk_size, :]).mean(axis=0)
         if single_class_mode:
             arr1 = sigmoid(arr1)
-
+        else:
```
remtav, line 225: Thanks for this fix

inference_segmentation.py
```diff
@@ -349,13 +335,12 @@ def main(params: Union[DictConfig, dict]) -> None:
     models_dir.mkdir(exist_ok=True)
     data_dir = get_key_def('raw_data_dir', params['dataset'], default="data", to_path=True, validate_path_exists=True)
     download_data = get_key_def('download_data', params['inference'], default=False, expected_type=bool)
+    override = get_key_def('override_model_params', params['inference'], default=False, expected_type=bool)
```
remtav, line 338: Cool

tests/utils/test_inference.py
remtav, line 1: I tried the added tests. They look good and make sense. Thanks.

utils/inference.py
```diff
@@ -49,3 +51,100 @@ def class_from_heatmap(heatmap_arr: np.ndarray, heatmap_threshold: float = 0.5)
...
+        window_size (int): Same as Chunk/Tile size used at inference.
+
+    Returns:
+        ndarray (2D): Signal window with values ranging from 0 to 1.
+    """
+    window = np.matrix(w.hann(M=window_size, sym=False))
+    return window.T.dot(window)
+
+
+def generate_corner_windows(window_size: int) -> np.ndarray:
```
remtav, line 70: Is there any article that could be cited as reference iin the docstrings?

utils/inference.py
```diff
@@ -49,3 +51,100 @@ def class_from_heatmap(heatmap_arr: np.ndarray, heatmap_threshold: float = 0.5)
...
+    """
+    patch_list = []
+    if overlapping:
+        step = window_size >> 1
+        windows = generate_corner_windows(window_size)
+        max_height = int(image_height/step - 1)*step
+        max_width = int(image_width/step - 1)*step
+    else:
+        step = window_size
+        windows = np.ones((window_size, window_size))
```
remtav, line 122: So, if no overlap, the hann window is pretty much bypassed, right? I hadn't realized, but that makes sense.

remtav (approved):

inference_segmentation.py
valhassan: I understand this, for next time I would keep this in mind to keep things separated better. I added all the small tweaks to improve prediction performance however marginal.

inference_segmentation.py
valhassan: Yes, agreed!

utils/inference.py
valhassan: Yes,

PR #496, opened 2023-03-08
change line in evaluate train seg

### LINKED ISSUE #493: multiclass loss computation error for validation and test
This [line](https://github.com/NRCan/geo-deep-learning/blob/develop/train_segmentation.py#L441) in the evaluation function of training pipeline gives error for multiclass loss calculation for losses such as cross entropy.
It should change to 
`loss = criterion(outputs, labels) if num_classes > 1 else criterion(outputs, labels.unsqueeze(1).float())`
The same way as it is computed in the training function.

CharlesAuthier:
fix #493 and add test for losses

### DIFF
--- a/train_segmentation.py
+++ b/train_segmentation.py
@@ -438,7 +438,7 @@ def evaluation(eval_loader,
                                    ep_num=ep_idx + 1,
                                    scale=scale)
 
-            loss = criterion(outputs, labels.unsqueeze(1).float())
+            loss = criterion(outputs, labels) if num_classes > 1 else criterion(outputs, labels.unsqueeze(1).float())
 
             eval_metrics['loss'].update(loss.item(), batch_size)
 
--- a/tests/loss/test_losses.py
+++ b/tests/loss/test_losses.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 
+import torch
 from hydra import initialize, compose
 from hydra.core.hydra_config import HydraConfig
 from hydra.utils import to_absolute_path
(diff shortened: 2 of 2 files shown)


train_segmentation.py
```diff
@@ -438,7 +438,7 @@ def evaluation(eval_loader,
                                    ep_num=ep_idx + 1,
                                    scale=scale)
 
-            loss = criterion(outputs, labels.unsqueeze(1).float())
+            loss = criterion(outputs, labels) if num_classes > 1 else criterion(outputs, labels.unsqueeze(1).float())
```
ms888ekb, line 441: have you tested that this line fails for the multiclass evaluation? 
In my tests loss = criterion(outputs, labels.unsqueeze(1).float()) works for both cases

CharlesAuthier: yeah I created a test for it, it dont crash for all binary but some yeah and it pass for the news tests for binary and multiclass loss function that have a yaml in config

config/dataset/test_ci_segmentation_binary.yaml
```diff
@@ -15,4 +15,4 @@ dataset:
   class_name: # will follow in the next version
   classes_dict: {'BUIL':1}
   class_weights:
-  ignore_index: 255
+  ignore_index: -1
```
ms888ekb: why ignore_index = -1? it is  a convention to use 255 for 8-bit data.

CharlesAuthier: yeah but it was like that before so I rechange it to the original value

ms888ekb: I think that 255 is the correct way to mark ignor_index pixels.
Let's hear what other mates think

ms888ekb (approved):

valhassan (approved):

PR #498, opened 2023-03-09
Fixe the cover of the tests already in place

### LINKED ISSUE #497: Test don't cover all the functions in test file
cover

In some test file we can see that the cover column is not 100% meaning that the test function don't go trough all the line wanted.

CharlesAuthier:
Fix #497, by changing the path of some `for` loop and removing some if-else statements to upgrade the cover to 100% (or close).

### DIFF
--- a/tests/dataset/test_aoi.py
+++ b/tests/dataset/test_aoi.py
@@ -203,6 +203,7 @@ class Test_AOI(object):
         data = read_csv("tests/tiling/tiling_segmentation_binary-multiband_ci.csv")
         row = next(iter(data))
         row['split'] = "missing_split"
+        print('AOI')
         with pytest.raises(ValueError):
             print('AOI')
             aoi = AOI(raster=row['tif'], label=row['gpkg'], split=row['split'])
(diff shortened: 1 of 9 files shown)


tests/dataset/test_aoi.py
```diff
@@ -114,10 +116,7 @@ def test_stac_url_input(self):
         assert aoi.download_data is True
         assert Path("data/SpaceNet_AOI_2_Las_Vegas-056155973080_01_P001-WV03-R.tif").is_file()
         aoi.close_raster()
-        try:
-            os.remove("data/SpaceNet_AOI_2_Las_Vegas-056155973080_01_P001-WV03-R.tif")
-        except PermissionError:
-            pass
+        os.remove("data/SpaceNet_AOI_2_Las_Vegas-056155973080_01_P001-WV03-R.tif")
```
ms888ekb, line 119: I think this kind of exceptions could be left unchanged because it really helps when testing locally under Windows OS

mpelchat04 (approved):

mpelchat04 (approved):

remtav (approved):

PR #504, opened 2023-03-28
Add Tag with classes dict to inference raster

CharlesAuthier:
Adding a tag containing the `classes_dict` parameter, find in the dataset yaml, to the output raster after inference.
Also adding a tests to verify the output,  **important** to use the `fetch_tag_raster` function to fetch the `classes_dict`, otherwise the output will be a string and not a dictionary.


remtav (approved):
Looks good to me!

ms888ekb (approved):

PR #507, opened 2023-04-04
Add a multilayer reading option for AOI

CharlesAuthier:
Add the option to read a multilayers GPKG when is read by the AOI function.

Plus tests for the entire `check_gdf_load` function.

### DIFF
--- a/utils/geoutils.py
+++ b/utils/geoutils.py
@@ -6,6 +6,7 @@ from typing import List, Union, Sequence
 
 import ast
 import pyproj
+import fiona
 from fiona._err import CPLE_OpenFailedError
 from fiona.errors import DriverError
 import geopandas as gpd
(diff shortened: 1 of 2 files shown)


utils/geoutils.py
```diff
@@ -174,11 +176,24 @@ def check_rasterio_im_load(im):
         raise ValueError("{} is not an accepted image format for rasterio.".format(im))
 
 
-def check_gdf_load(gdf):
+def check_gdf_load(gdf: Union[str, Path, gpd.GeoDataFrame]) -> gpd.GeoDataFrame:
```
ms888ekb, line 179: good job Charles! I like how you improved this part.

utils/geoutils.py
```diff
@@ -190,7 +205,21 @@ def check_gdf_load(gdf):
...
+                for count, layername in enumerate(layer_name):
+                    if count == 0: 
+                        multi_layers = gpd.read_file(gdf, layer=layername)
+                    else: 
+                        geopkg = gpd.read_file(gdf, layer=layername)
+                        multi_layers = pd.concat(
+                            [multi_layers, geopkg],
+                            ignore_index=True
+                        )    
+                return multi_layers
```
ms888ekb, line 222: In general, I understand why it makes sense to know if the label gpkg is a multi-layer or not. But do we have label gpkg in practice? I thought we always have a single-layer single-class gpkg for training? If this is not true, then I need to adjust my tiling script too.

CharlesAuthier: We mustly use single-layer single-class, but we want to also support multi-classes and Pierre told me that the GPKG will be multi layer.

PR #526, opened 2023-05-24
geoutils.py: force BIGTIFF

### LINKED ISSUE #525: BUG: GDL cannot write TIFFs larger than 4 GB
**Description**
GDL writes tifs at inference and these tifs may become very large:
- inference: [enhanced version of imagery](https://github.com/NRCan/geo-deep-learning/blob/0e09cb1150cb05169eee316b213ec280646f1817/dataset/aoi.py#LL473C23-L473C23)
- inference: [raw inference heatmap]((https://github.com/NRCan/geo-deep-learning/blob/develop/inference_segmentation.py#L464))
- inference: [raw inference after argmax/sigmoid of the heatmap array](https://github.com/NRCan/geo-deep-learning/blob/develop/inference_segmentation.py#L477)

**To Reproduce**
Steps to reproduce the behavior on HPC: see `[path/to]/operationalization_dev/geosys-jobs/jobs/NB_DNR_2013_a_INF_BUIL_2023-05-24T09h38_inference.job`
This inference uses a huge image as input which results in the following error when it's time to write the inference to tif.

This results in the following error:
`GDAL signalled an error: err_no=1, msg='TIFFAppendToStrip:Maximum TIFF file size exceeded. Use BIGTIFF=YES creation option.'`

**Expected behavior**
We want to be able to write TIFFs larger than 4 gb. This can be done by modifying our [`create_new_raster_from_base`](https://github.com/NRCan/geo-deep-learning/blob/develop/utils/geoutils.py#L28) to use the BIGTIFF=YES as suggested [here](https://gis.stackexchange.com/questions/347765/how-to-write-bigtiff-raster-files-with-rasterio).

remtav:
fixes #525 

In the [GDAL doc](https://gdal.org/drivers/raster/gtiff.html), nothing seems problematic with always writing BIGTIFF. The only requirement is to have libtiff > 4.0, which has also been the case in our conda envs. I check all our conda envs and docker images and they all include libtiff >4.0.

This solution also uses tiled=True and defines blocksizes in the tif, as suggested by [this SO post](https://gis.stackexchange.com/questions/347765/how-to-write-bigtiff-raster-files-with-rasterio).

### DIFF
--- a/inference_segmentation.py
+++ b/inference_segmentation.py
@@ -172,7 +172,7 @@ def segmentation(param,
     img_gen = gen_img_samples(src=input_image, patch_list=patch_list, chunk_size=chunk_size)
     single_class_mode = False if num_classes > 1 else True
     for sub_image, h_idxs, w_idxs, hann_win in tqdm(
-        img_gen, position=0, leave=True, desc='Inferring on patches',
+        img_gen, position=0, leave=True, desc=f'Inferring on patches',
         total=len(patch_list)
     ):
         hann_win = np.expand_dims(hann_win, -1)
--- a/inference_segmentation.py
+++ b/inference_segmentation.py
@@ -172,7 +172,7 @@ def segmentation(param,
     img_gen = gen_img_samples(src=input_image, patch_list=patch_list, chunk_size=chunk_size)
     single_class_mode = False if num_classes > 1 else True
     for sub_image, h_idxs, w_idxs, hann_win in tqdm(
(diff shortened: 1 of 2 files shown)


remtav:
Having trouble with my rebasing. Lots of duplicate commits. They will be squashed anyways but I'll try to handle things better next time...

CharlesAuthier (approved):

mpelchat04 (approved):

ms888ekb (approved):

Issue #514, opened 2023-05-10
add PR template


remtav:
A PR template would ensure contributors follow a uniform procedure when opening PRs
[PR template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository)
[Example](https://github.com/CosmiQ/solaris/blob/main/PULL_REQUEST_TEMPLATE.md)

valhassan:
Another example: https://www.talater.com/open-source-templates/#/page/1

LucaRom:
> Another example: https://www.talater.com/open-source-templates/#/page/1

I enjoyed the adventure!

PR #530, opened 2023-06-05
Fix gh actions

ms888ekb:
fix for GitHub Action tests


CharlesAuthier:
Can you remove what you have comment plz
I dont like the fact that we need to switch a lot of the package from conda to pip, but if its what it take I accept it.

Thanks a lot, I was struggling whit that. 

And can you add  "Fix #514" to your description plz

ms888ekb:
> Can you remove what you have comment plz I dont like the fact that we need to switch a lot of the package from conda to pip, but if its what it take I accept it.
> 
> Thanks a lot, I was struggling whit that.

I'll take a look later which packages can be moved back to conda but for now this combination works

CharlesAuthier (approved):

mpelchat04 (approved):

PR #548, opened 2023-09-29
Hot fix docker image build

ms888ekb:
pointed system variable from the OS's to Conda's installations lib folder.


CharlesAuthier:
can you combine all the commit to make it clean pls :)

valhassan (approved):
Removing the container function seems like a temporary fix. Ideally, we would like to run our tests in the same environment (docker container) as we would in production. This will remove any validation of the docker image. Also, I think all docker image builds should be done inside of GitHub actions (Deprecating build-on-tag). We need more discussions on this!

remtav (approved):

